Issue #128, opened 2022-01-25
TASK: Implement Scaffolding


russellmorton:
### Description

As a user, I want to initialize a new project using the smart contract language, and, optionally, a web framework of my choice.

### Design notes

`taq scaffold [alias|uri]`

#### Alias examples

The following respecitve commands would scaffold a project with a basic smart contract, along with (optionally) tests in the SC's language. A deployment script and example integration test will also be generated by the scaffold action.

`taq scaffold ligo`
`taq scaffold smartpy`

#### URI example

A third party can make a scaffold and push it to a git server. 

Any user of `taq` should be able to use that scaffold by running the command;

`taq scaffold https://github.com/example/tezos-ligo-and-react-nft-market`

Which would give the user a scaffold-ed project with ligo NFT market place contracts, a started dapp, deployment and test scripts.

### Acceptance Criteria

- [ ] `quick-start` repo is created and can be cloned
- [ ] `quick-start` is the default github repo
- [ ] `quick-start` is the default target directory

ricklove:
@mweichert I think the [type] argument is beyond the scope of the most basic implementation. For the minimal implementation, it just needs the git url and target directory.

PR #139, opened 2022-01-26
Project Scaffold

ricklove:
- [x] Run git clone of the scaffoldUrl (default: https://github.com/ecadlabs/taqueria-scaffold-quickstart)
- [x] Cleanup (delete .taq/scaffold.json & .git)
- [x] Create quick-start scaffold project: https://github.com/ecadlabs/taqueria-scaffold-quickstart
- [ ] ~Run an initialize command~ (this can be provided by readme for initial implementation)

## Related Issues:
- https://github.com/ecadlabs/taqueria/issues/128

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -1,14 +1,14 @@
-import type {UnvalidatedPluginInfo, InstalledPlugin, Config, ConfigArgs, Action, Alias, Verb, Option, PositionalArg} from './taqueria-protocol/taqueria-protocol-types.ts'
+import type {UnvalidatedPluginInfo, InstalledPlugin, Config, ConfigArgs, Alias, Verb, Option, PositionalArg, PluginResponse, PluginAction, ProxyAction} from './taqueria-protocol/taqueria-protocol-types.ts'
 import {Task, PluginInfo} from './taqueria-protocol/taqueria-protocol-types.ts'
-import type {EnvKey, EnvVars, DenoArgs, RawInitArgs, SanitizedInitArgs, i18n, InstallPluginArgs, UninstallPluginArgs, SanitizedScaffoldInitArgs} from './taqueria-types.ts'
+import type {EnvKey, EnvVars, DenoArgs, RawInitArgs, SanitizedInitArgs, i18n, InstallPluginArgs, UninstallPluginArgs} from './taqueria-types.ts'
 import {State} from './taqueria-types.ts'
 import type {Arguments} from 'https://deno.land/x/yargs/deno-types.ts'
 import yargs from 'https://deno.land/x/yargs/deno.ts'
-import {map, chain, attemptP, chainRej, resolve, reject, fork, forkCatch, parallel, debugMode, attempt} from 'https://cdn.jsdelivr.net/gh/fluture-js/Fluture@14.0.0/dist/module.js'
+import {map, chain, attemptP, mapRej, chainRej, resolve, reject, fork, forkCatch, parallel, debugMode} from 'https://cdn.jsdelivr.net/gh/fluture-js/Fluture@14.0.0/dist/module.js';
 import {pipe, identity} from "https://deno.land/x/fun@v1.0.0/fns.ts"
 import {getConfig, getDefaultMaxConcurrency} from './taqueria-config.ts'
-import {as, ensurePathDoesNotExist, exec, gitClone, isTaqError, joinPaths, mkdir, readJsonFile, rm, writeTextFile} from './taqueria-utils/taqueria-utils.ts'
-import {SanitizedAbsPath, SanitizedPath, TaqError, Future, SanitizedUrl} from './taqueria-utils/taqueria-utils-types.ts'
+import {exec, isTaqError, joinPaths, mkdir, readJsonFile, writeTextFile, decodeJson, ensurePathDoesNotExist, gitClone, rm} from './taqueria-utils/taqueria-utils.ts'
+import {SanitizedAbsPath, SanitizedPath, SanitizedUrl, TaqError, Future} from './taqueria-utils/taqueria-utils-types.ts'
 import {Table} from 'https://deno.land/x/cliffy@v0.20.1/table/mod.ts'
 import { titleCase } from "https://deno.land/x/case/mod.ts";
 import {uniq} from 'https://deno.land/x/ramda@v0.27.2/mod.ts'
@@ -126,7 +126,7 @@ const commonCLI = (env:EnvVars, args:DenoArgs, i18n: i18n) =>
             ({projectDir, configDir, maxConcurrency, quickstart}: SanitizedInitArgs) => {
                 return initProject(projectDir, configDir, i18n, maxConcurrency, quickstart)
             },
-            fork (console.error) (console.log)
+            forkCatch (console.error) (console.error) (console.log)
         )
     )
     .command(
@@ -146,11 +146,9 @@ const commonCLI = (env:EnvVars, args:DenoArgs, i18n: i18n) =>
                 })
         },
         (args: RawInitArgs) => pipe(
-            sanitizeScaffoldArgs(args), 
-            ({scaffoldUrl, scaffoldProjectDir, configDir, maxConcurrency, quickstart}: SanitizedScaffoldInitArgs) => {
-                return scaffoldProject(scaffoldUrl, scaffoldProjectDir, configDir, i18n, maxConcurrency, quickstart)
-            },
-            fork (console.error) (console.log)
+            sanitizeArgs(args),
+            scaffoldProject(i18n),
+            forkCatch (console.error) (console.error) (console.log)
         )
     )
     .option('fromVsCode', {
@@ -188,7 +186,7 @@ const postInitCLI = (cliConfig: CLIConfig, env: EnvVars, args: DenoArgs, parsedA
         // a plugin distributed and installable via NPM. This should support other means of distribution
         (inputArgs: InstallPluginArgs) => pipe(
             NPM.installPlugin(parsedArgs.configDir, parsedArgs.projectDir, i18n, inputArgs.pluginName),
-            fork (displayError(cliConfig)) (console.log)
+            forkCatch (displayError(cliConfig)) (displayError(cliConfig)) (console.log)
         )
     )
     .alias('i', 'install')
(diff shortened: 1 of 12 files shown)


cli.ts
```diff
@@ -294,6 +319,66 @@ const initProject = (projectDir: SanitizedAbsPath, configDir: SanitizedPath, i18
...
+        console.log(`initializing...`)
+        // TODO: Run initialization script
+        // Load .taq/scaffold.json (if it exists)
+        // Run init command
+
+
+        // Cleanup:
+        console.log(`cleanup...`)
+
+        try {
```
mweichert, line 361: Another good example of a reusable utility here. Would you be up for adding a `rm` utility to taqueria-utils with a signature something like:
`string => Future<E_RM, SanitizedAbsPath>` ?

cli.ts
```diff
@@ -294,6 +319,66 @@ const initProject = (projectDir: SanitizedAbsPath, configDir: SanitizedPath, i18
...
+            console.log(`removing ${joinPaths(projectDir.value, `.taq/scaffold.json`)}`)
+            await Deno.remove(joinPaths(projectDir.value, `.taq/scaffold.json`));
+        } catch {
+            // Ignore if doesn't exist
+        }
+        try {
```
mweichert, line 367: Glad to see that you're removing the .git directory here. Good thought!

Probably obvious, but if we're adding the `rm` utility mentioned earlier we'd want to use it in this case as well.

cli.ts
```diff
@@ -740,6 +825,8 @@ const sanitizeArgs = (parsedArgs: RawInitArgs) : SanitizedInitArgs => ({
     _: parsedArgs._,
     configDir: SanitizedPath.create(parsedArgs.configDir),
     projectDir: SanitizedAbsPath.create(parsedArgs.projectDir),
+    // TODO: SanatizedUrl
```
mweichert, line 828: 👍

mweichert (changes requested):
Thanks so much for your work here, Rick.

I hope you don't mind, but I did add a few suggestions - most of them taking some of the processes that are executed in the scaffold task and moving them to reusable utility functions.

Excellent work, and much appreciated.

Btw, let me know if you'd like to have a chat about Futures and how to chain/map over them.

cli.ts
```diff
@@ -126,6 +126,31 @@ const commonCLI = (env:EnvVars, args:DenoArgs, i18n: i18n) =>
...
+        'scaffold [scaffoldUrl]',
+        i18n.__('scaffoldDesc'),
+        (yargs: Arguments) => {
+            yargs
+                .positional('scaffoldUrl', {
+                    describe: i18n.__('scaffoldUrlDesc'),
+                    type: 'string',
+                    default: 'https://github.com/ecadlabs/taqueria-scaffold-quickstart.git',
+                })
+                .option('p', {
```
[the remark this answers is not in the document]
ricklove: Yes, this is same option - I was just overriding the description for that option. 

Is this the right way to do that?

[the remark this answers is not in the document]
mweichert: OK, sounds good. Perhaps we could call that positional arg destDir to avoid a naming collision with projectDir. Thanks!

cli.ts
```diff
@@ -294,6 +319,66 @@ const initProject = (projectDir: SanitizedAbsPath, configDir: SanitizedPath, i18
...
+            await Deno.stat(projectDir.value);
+            return Promise.reject({kind: 'E_SCAFFOLD_PROJECT_DIR_ALREADY_EXISTS', msg: 'TODO i18n message'})
+        } catch(_e) {
+            // Expect exception when trying to stat a new directory
+        }
+
+        // git clone
+        console.log(`git clone...`)
+
+        const cloneProcess = Deno.run({
```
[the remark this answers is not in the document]
ricklove: Ok, I'll break this into steps. I intended that originally, but I wasn't sure how to work with Future. Thanks for the suggestion on the type - that will help me get started.

cli.ts
```diff
@@ -1,14 +1,14 @@
 import type {UnvalidatedPluginInfo, InstalledPlugin, Config, ConfigArgs, Alias, Verb, Option, PositionalArg, PluginResponse, PluginAction, ProxyAction} from './taqueria-protocol/taqueria-protocol-types.ts'
 import {Task, PluginInfo} from './taqueria-protocol/taqueria-protocol-types.ts'
-import type {EnvKey, EnvVars, DenoArgs, RawInitArgs, SanitizedInitArgs, i18n, InstallPluginArgs, UninstallPluginArgs} from './taqueria-types.ts'
+import type {EnvKey, EnvVars, DenoArgs, RawInitArgs, SanitizedInitArgs, i18n, InstallPluginArgs, UninstallPluginArgs, SanitizedScaffoldInitArgs} from './taqueria-types.ts'
```
mweichert, line 3: The name of `RawInitArgs` and `SanitizedInitArgs` is misleading. I should have named them 'RawArgs' and 'SanitizedArgs', as they aren't specific to the `init` task. To be consistent with the other named types, I should have probably named these "UnvalidedTaskArgs" and "TaskArgs" - that would follow the naming convention used elsewhere.

In any case, what's your thoughts on just consolidating SanitizedInitArgs and SanitizedScaffoldInitArgs?

Although having separate types for each built-in task might seem appealing, it might just add noise - yargs parsing of ALL args is done atomically, so my thinking here is that we just need a type to represent the args after they were parsed by yargs, and then another type to represent those args after we've done validation and sanitization for our limited set of built-in tasks. 

I could probably go either way on this though, so I'd be curious what your opinion is. If we do think having separate types for sanitized args specific to each built-in task makes sense, then I should probably refactor and segment SanitizedInitArgs, as right now the intent was that this type would be a representation of sanitized args after yargs parsed them, and that these args are valid and can be used by any built-in task.

ricklove: The reason I had to separate those is because of the difference between optional types:

```
export interface SanitizedInitArgs {
    scaffoldUrl?: SanitizedUrl;
    scaffoldProjectDir?: SanitizedAbsPath;
}
export interface SanitizedScaffoldInitArgs extends SanitizedInitArgs {
    scaffoldUrl: SanitizedUrl;
    scaffoldProjectDir: SanitizedAbsPath;
}
const sanitizeScaffoldArgs = (parsedArgs: RawInitArgs) : SanitizedScaffoldInitArgs => ({
    ...sanitizeArgs(parsedArgs),
     scaffoldUrl: SanitizedUrl.create(parsedArgs.scaffoldUrl ?? ''),
     scaffoldProjectDir: SanitizedAbsPath.create(parsedArgs.scaffoldProjectDir ?? ''),
})
```

Only the `sanitizeScaffoldArgs` can guarantee that those values actually exist.

Any thoughts on a better pattern here?

cli.ts
```diff
@@ -129,6 +129,31 @@ const commonCLI = (env:EnvVars, args:DenoArgs, i18n: i18n) =>
...
+                    describe: i18n.__('scaffoldProjectDirDesc'),
+                    default: './taqueria-quickstart',
+                })
+        },
+        (args: RawInitArgs) => pipe(
+            sanitizeScaffoldArgs(args), 
+            ({scaffoldUrl, scaffoldProjectDir, configDir, maxConcurrency, quickstart}: SanitizedScaffoldInitArgs) => {
+                return scaffoldProject(scaffoldUrl, scaffoldProjectDir, configDir, i18n, maxConcurrency, quickstart)
+            },
+            fork (console.error) (console.log)
```
mweichert, line 154: This is new - in one of my latest branches, I switched all uses of `fork` to `forkCatch`. You can read about the differences here:
https://github.com/fluture-js/Fluture#fork

TLTR excerpt from their docs:
> If an exception was encountered during the computation, it will be re-thrown by fork and likely not be catchable. You can handle it using process.on('uncaughtException') in Node, or use [forkCatch](https://github.com/fluture-js/Fluture#forkcatch).

cli.ts
```diff
@@ -705,6 +748,11 @@ const sanitizeArgs = (parsedArgs: RawInitArgs) : SanitizedInitArgs => ({
     setVersion: parsedArgs.setVersion,
     version: parsedArgs.version
 })
+const sanitizeScaffoldArgs = (parsedArgs: RawInitArgs) : SanitizedScaffoldInitArgs => ({
```
mweichert, line 751: Relates to https://github.com/ecadlabs/taqueria/pull/139/files#r810054510

cli.ts
```diff
@@ -129,6 +129,31 @@ const commonCLI = (env:EnvVars, args:DenoArgs, i18n: i18n) =>
...
+                .positional('scaffoldProjectDir', {
+                    alias: 'scaffoldProjectDir',
+                    type: 'string',
+                    describe: i18n.__('scaffoldProjectDirDesc'),
+                    default: './taqueria-quickstart',
+                })
+        },
+        (args: RawInitArgs) => pipe(
+            sanitizeScaffoldArgs(args), 
+            ({scaffoldUrl, scaffoldProjectDir, configDir, maxConcurrency, quickstart}: SanitizedScaffoldInitArgs) => {
```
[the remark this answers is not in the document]
ricklove: In this case, I was copying the style used in the block above (and I just noticed some unneccessary parameters)

What are your thoughts on this:

```
        (args: RawInitArgs) => pipe(
            sanitizeScaffoldArgs(args), 
            (args: SanitizedScaffoldInitArgs) => scaffoldProject({...args, i18n } ),
            fork (console.error) (console.log)
        )
        ...
        
const scaffoldProject = ({
    scaffoldUrl,
    scaffoldProjectDir,
    i18n,
}: {
    scaffoldUrl: SanitizedUrl, 
    scaffoldProjectDir: SanitizedAbsPath, 
    i18n: i18n, 
}) => pipe(        
```

cli.ts
ricklove: Ok, will comment there

taqueria-utils/taqueria-utils.ts
```diff
@@ -19,19 +19,64 @@ export const decodeJson = <T>(encoded: string): Future<TaqError, T> =>
...
 const mkdirFuture = (path: string): Future<TaqError, void> => attemptP(() => Deno.mkdir(path, {recursive: true}))
+export const mkdir = (path: string) : Future<TaqError, string> => pipe(path, mkdirFuture, map(() => path))
 
-export const mkdir = (path: string) : Future<TaqError, string> => pipe(path, mkdirFuture, map (() => path))
+export const ensurePathExists = (path: string) : Future<TaqError, SanitizedAbsPath> => attemptP(async () =>{
+    try {
+        await Deno.stat(path);
+        return SanitizedAbsPath.create(path);
+    } catch(_e) {
+        return Promise.reject({ kind: 'E_INVALID_PATH_DOES_NOT_EXIST', msg: 'TODO i18n message' })
```
mweichert, line 44: I've removed my use of "TODO i18n message" - I used it in so many places that when the error was thrown, I couldn't determine where the error actually occurred. I think you're trying to follow what I've done in the past here, but it turned out to be a bad idea. Perhaps we could use this instead:

```
catch (previous) {
  // TODO: i18n error message
  return Promise.reject({ kind: 'E_INVALID_PATH_DOES_NOT_EXIST', 
    msg: 'The path doesn't exist', context: path, previous })
}
```

taqueria-utils/taqueria-utils.ts
```diff
@@ -19,19 +19,64 @@ export const decodeJson = <T>(encoded: string): Future<TaqError, T> =>
...
+    }  
+});
+
+export const ensurePathDoesNotExist = (path: string) : Future<TaqError, SanitizedAbsPath> => attemptP(async () =>{
+    try {
+        await Deno.stat(path);
+        return Promise.reject({ kind: 'E_INVALID_PATH_ALREADY_EXISTS', msg: 'TODO i18n message' })
```
mweichert, line 51: Same thoughts as my previous comment:
https://github.com/ecadlabs/taqueria/pull/139/files#r810071895

taqueria-utils/taqueria-utils.ts
```diff
@@ -19,19 +19,64 @@ export const decodeJson = <T>(encoded: string): Future<TaqError, T> =>
...
+
+export const gitClone = (url: SanitizedUrl) => (destinationPath: SanitizedAbsPath) => pipe(
+    attemptP(async () => {
+        const cloneProcess = Deno.run({
+            cmd: ["git", "clone", url.value, destinationPath.value],
+        });
+        const cloneResult = await cloneProcess.status();
+        
+        if (!cloneResult.success) {
+            return Promise.reject({kind: 'E_SCAFFOLD_URL_GIT_CLONE_FAILED', msg: 'TODO i18n message'})
```
mweichert, line 75: Related to a previous comment:
https://github.com/ecadlabs/taqueria/pull/139/files#r810071895

cli.ts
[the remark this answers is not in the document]
ricklove: Ah point-free style, no I see:

```
        (args: RawInitArgs) => pipe(
            sanitizeScaffoldArgs(args), 
            scaffoldProject(i18n),
            fork (console.error) (console.log)
        )

const scaffoldProject = (i18n: I18N) => ({
    scaffoldUrl,
    scaffoldProjectDir,
}: SanitizedScaffoldInitArgs) => pipe(
```

mweichert (changes requested):
Looking good Rick.

There were just a couple of small things that require either adjustment or feedback.

taqueria-utils/taqueria-utils.ts
ricklove: I'll search for any `TODO i18n message` and fix those

ricklove:
@mweichert I put some changes in a separate PR because I wanted your feedback first:
https://github.com/ecadlabs/taqueria/pull/336

Issue #38, opened 2021-12-10
TASK - Finalize CI Pipeline


russellmorton:
### Description
Finalize CI Pipeline to ensure the ability to test any build.

#### Acceptance Criteria
When Pull request is open/updated:
- [x] Build and install Taqueria
- [x] Install all required Taqueria plugins if needed
- [x] Run unit tests, integration and end-to-end tests
- [x] Pull request must contain binary
- [x] Publish to NPM

#### Additional Context and Related Items
https://github.com/ecadlabs/taquito/issues/506

PR #146, opened 2022-01-27
ci(docker): Add workflow to build and publish images

GImbrailo:
The PR sets up the docker build and publish workflow for a list of plugins in a matrix. This is only needed for the `flextesa` plugin currently

**Test Plan**
Please see the Release section of the repo for listed packages

image

image

Please see this PR for the docker comments
 
Ref #38

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v2
         with:
-          context: .
+          context: ./
           push: true
           tags: ${{ steps.image-metadata.outputs.tags }}
           labels: ${{ steps.image-metadata.outputs.labels }}
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v2
         with:
-          context: ./
+          context: ${{ matrix.plugin }}/docker
           push: true
           tags: ${{ steps.image-metadata.outputs.tags }}
           labels: ${{ steps.image-metadata.outputs.labels }}
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -27,7 +27,7 @@ jobs:
           - plugin: taqueria-plugin-flextesa 
     defaults:
       run:
-        working-directory: ${{ matrix.plugin }}
+        working-directory: ${{ matrix.plugin }}/docker
 
     steps:
       - uses: actions/checkout@v2
@@ -53,7 +53,7 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v2
         with:
-          context: ${{ matrix.plugin }}/docker/
+          context: .
           push: true
           tags: ${{ steps.image-metadata.outputs.tags }}
           labels: ${{ steps.image-metadata.outputs.labels }}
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,67 @@
...
+
+jobs:
+
+  build-and-push-docker:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - plugin: taqueria-plugin-flextesa
```
GImbrailo, line 26: I think it is useful to run this job as a matrix. This way we can add any new plugins that require docker builds to the matrix. I am happy to remove it for now if needed

taqueria-plugin-flextesa/docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM oxheadalpha/flextesa:latest
 RUN mkdir /app
 WORKDIR /app
-COPY startFlextesa.js /app/
+COPY startFlextesa.ts /app/
 RUN apk add net-tools && apk add nodejs && apk add npm
```
jevonearth, line 5: Running `RUN apk add net-tools nodejs npm` will complete faster.

jevonearth (approved):

jevonearth (approved):

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,67 @@
...
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+    branches:
+      - 'main'
+    paths:
+      - 'taqueria-plugin-*/docker/*'
+  create:
```
danielelisi, line 12: I think we should use `push: tags` here like the other workflow

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,67 @@
...
+    tags:
+      - 'v*'
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+
+  build-and-push-docker:
```
danielelisi, line 20: You can remove the docker suffix since this job is already in the docker workflow

danielelisi (commented):
How does this pipeline detect when to push a semver tag? From what I read seems like anyone could push a semver tags from a pr?

.github/workflows/docker-publish.yml
GImbrailo: :upside_down_face: Changed

.github/workflows/docker-publish.yml
GImbrailo: That makes sense. Changed

GImbrailo:
> How does this pipeline detect when to push a semver tag? From what I read seems like anyone could push a semver tags from a pr?

@danielelisi from my understanding this is done in the `Exctract metadata` job. The tags are set and defined on lines 45 -49 of the workflow

```yaml
          tags: |
            type=ref,event=branch
            type=ref,event=pr,suffix=-{{sha}}
            type=semver,pattern={{version}}
            type=semver,pattern={{major}}.{{minor}}
```

The documentation on the [docker/metadata-action](https://github.com/docker/metadata-action) repo provide more information on the various types and how to configure them

GImbrailo:
@danielelisi based on the comments in the SRE standup, I am going to make some more changes here

.github/workflows/docker-publish.yml
GImbrailo: As discussed in the SRE standup the `flextesa` docker image is an outlier. I will remove the matrix configuration

danielelisi (approved):

PR #149, opened 2022-01-27
Fixes #147, tasks not exposed in cli - among other changes

### LINKED ISSUE #147: Fixed problem with tasks not being listed since renaming of packages
### Description

Recently, we renamed all NPM packages. For instance, taqueria-plugin-ligo became @taqueria/plugin-ligo.

Unfortunately, this change broke a few different things:

#### Root README.md needs updated

The documentation in the root README.md instructs to modify the .taq/config.json file by adding an InstalledPlugin definition to the plugins array. The pattern used for the package name followed the old convention, such as taqueria-plugin-ligo. 

If an end-user were to follow these instructions and run `taq`, they'll see the following error:

```
{
  kind: "E_INVALID_JSON",
  msg: "TODO i18n message",
  context: {
    stderr: "node:internal/modules/cjs/loader:936\n  throw err;\n  ^\n\nError: Cannot find module '/Users/mweichert/P...",
    stdout: ""
  }
}
```

#### No tasks provided by an installed plugin are exposed in the CLI

If an end-user were to use the updated package name, such as @taqueria/plugin-ligo in their .taq/config.json file, they wouldn't get the error above, but instead, no tasks from that plugin will be listed in the CLI. This is a significant bug.

#### Acceptance Criteria
- [ ] Root README.md is updated with the correct instructions
- [ ] The CLI displays all tasks exposed by an installed plugin

#### Additional Context and Related Items

I've discovered that the code that needs adjustment is in the cli.ts file, and in particular, the loadState() function.

mweichert:
This branch fixes #147 but also makes a few other adjustments:
- Adds a task for installing and uninstalling an NPM plugin.  For installation, you can use `taq install @taqueria/plugin-ligo` once the packages are published to npmjs, but `taq install /path/to/taqueria-plugin-ligo` should also work. To uninstall, you must specify the package name - directory paths aren't supported.
- Error messages aren't truncated. This will help with debugging.
- Adds an argument flag for logging plugin requests. When debugging, I often have to uncomment a line in the source of taqueria and re-build. I can then see calls to a plugin logged to the console. I've asked others, like @russellmorton to do this. To make this a little easier, you can now just specify a hidden argument to the taq binary: `taq --logPluginCalls`
- When you execute a task using taqueria for the first time, we generate a .taq/state.json file which makes taqueria fast and performant. However, when debugging, you often have to delete this file in-between executions. To make this easier, you can now specifiy a hidden argument: `taq --disableState` which will generate a new state file for every execution. This makes taqueria slower, but should only be used for debugging anyhow.
- Its possible that we make changes to the schema used in state.json in later versions. Indeed, there's a slight schema change in this branch. The state.json file now includes a "build" property in the schema, which tracks the version of taqueria that generated the state.json file. Taqueria will now detect if the version of taqueria that you're using differs from the one that generated the state.json file, and regenerates it to assure that its using the most up-to-date schema
- One of the things we want to be able to do is have taqueria report what build (branch, commit, etc) was used to generate the binary. Although I haven't added that, I did add a mechanism that can be used to provide that information to taqueria. The next step is now just displaying/reporting that.
- In a call with Russel, we found that you cannot run `npm install [pluginName]` on a taqified directory if the directory isn't configured as an NPM project. This is done using the `npm init` command. Taqueria will now identify if this is an NPM project, and prompt you to run `npm init` if you try installing a plugin before the directory has been initialized as an NPM project

I think that's it. Although the above may look insignificant, there were changes required in a few different places to accommodate everything, so testing this branch before merging into main would be highly desirable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,15 +26,6 @@ I like adding my project directory to the PATH environment variable which allows
 1. Initialize a new project: `taqueria init test-project`
 2. Change directories: `cd test-project`
 3. Initialize the project as an NPM project: `npm init -y`
-4. Install the LIGO plugin: `npm i -D ../taqueria-plugin-ligo`
-5. Activate the LIGO plugin by adding the following to the plugins array in ./.taq/config.json:
-```json
-{
-    "name": "taqueria-plugin-ligo",
-    "type": "npm"
-}
-```
-
-> NOTE: We will be implementing a `taqueria install` task which will do steps 3-5 for you.
+4. Install the LIGO plugin: `taq install ../taqueria-plugin-ligo` if installing from a clone of this repo, otherwise use `taq install @taqueria/plugin-ligo`
 6. Continue steps 4-5 for each additional plugin you want to install
 
--- a/cli.ts
+++ b/cli.ts
@@ -6,8 +6,8 @@ import type {Arguments} from 'https://deno.land/x/yargs/deno-types.ts'
 import yargs from 'https://deno.land/x/yargs/deno.ts'
 import {map, chain, attemptP, chainRej, resolve, reject, fork, forkCatch, parallel, debugMode} from 'https://cdn.skypack.dev/fluture';
 import {pipe, identity} from "https://deno.land/x/fun@v1.0.0/fns.ts"
-import {getConfig, getDefaultMaxConcurrency} from './taqueria-config.ts'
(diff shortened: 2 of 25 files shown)


mweichert:
Collaboratively the team tested this one together and feel confident merging.

PR #153, opened 2022-01-27
fix(ci): assign exec permission to binaries

danielelisi:
## Summary

Currently after downloading a binary from Github we need to set exec permission ouself with `chmod +x`. Ideally we should let the pipeline do this for us.


GImbrailo (approved):

danielelisi:
According to this Issue https://github.com/actions/upload-artifact/issues/38 doesn't seem like we can do this unless we `tar` the file https://github.com/actions/upload-artifact#maintaining-file-permissions-and-case-sensitive-files

I'll close this PR for now as we can just add a step in the docs to change file permission.

PR #156, opened 2022-01-28
added reusable workflow for publishing packages

alexzbusko:
This PR makes the following changes:

- Breaks out the npm publishing job in the main workflow into a reusable workflow that is called multiple times for all the different npm packages
- Removes concurrency from the workflows due to it cancelling npm publish jobs (reference here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow)
- Updates the PR comment job to include the flextesa plugin
- Added a .npmignore file to the flextesa plugin to not publish a docker image to a npm package
- Updated package.json files for the plugins to have CI version of build and config updates to ensure both local building and CI building works


.github/workflows/main.yml
```diff
@@ -92,67 +85,68 @@ jobs:
...
+      dir: taqueria-plugin-taquito
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPMJS_PAT }}
+
+  publish-flextesa-plugin-to-npm:
+    needs:
+      - publish-ligo-plugin-to-npm
+      - publish-smartpy-plugin-to-npm
+      - publish-taquito-plugin-to-npm
+    uses: ecadlabs/taqueria/.github/workflows/npm-publish.yml@alex-npm-pre-build-ordering
```
GImbrailo, line 133: The branch name should be removed for these

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: npm ci
+
+      - name: Build package
+        run: npm run build-ci
+      
+      - name: Publish for PR
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          npm --no-git-tag-version version "0.0.0-pr-${{ github.event.number }}-${GITHUB_SHA::8}"
+          npm publish --tag "pr-${{ github.event.number }}" --access public
```
GImbrailo, line 55: It was recently decided that we should replace the `pr-#` tag with `preview` to prevent our npmjs.com tags from getting out of control

.github/workflows/main.yml
alexzbusko: @GImbrailo you're right but they need to be changed to `main`. From the docs I could find (https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow) this is the only way I could get them to call correctly. I tried to get it to call from the other workflow file but kept getting an error that made no sense to me. I think this is another peculiarity with github actions

.github/workflows/npm-publish.yml
alexzbusko: Okay cool, I will make the change

.github/workflows/main.yml
```diff
@@ -12,17 +12,10 @@ on:
   release:
     types: [published]
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref != 'refs/heads/main' || github.run_number }}
-  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
```
danielelisi, line 17: This makes sure that we only run pipelines for latest commits to a PR. Why are you removing it?

.github/workflows/main.yml
danielelisi: You could have also split this PR and have a separate one creating the reusable workflow so that you could merge them in order.

Anyways you can also remove the branch name right before merging to `main`

.github/workflows/main.yml
```diff
@@ -92,67 +85,68 @@ jobs:
...
+  publish-taquito-plugin-to-npm:
+    needs: publish-sdk-to-npm
+    uses: ecadlabs/taqueria/.github/workflows/npm-publish.yml@alex-npm-pre-build-ordering
+    with:
+      dir: taqueria-plugin-taquito
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPMJS_PAT }}
+
+  publish-flextesa-plugin-to-npm:
+    needs:
```
mweichert, line 129: @alexzbusko great work thus far! 

Small thing - the flextesa plugin doesn't need these plugins to be built first:
- publish-ligo-plugin-to-npm
- publish-smartpy-plugin-to-npm
- publish-taquito-plugin-to-npm

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,67 @@
...
+          echo "::set-output name=npm-tag::next"
+
+      - name: Prepare package.json for publishing
+        run: npm run prod-config-ci
+
+      - name: Install any dependencies
+        run: npm ci
+
+      - name: Build package
+        run: npm run build-ci
```
danielelisi, line 49: Why do we need a separate build script for CI ? I see most of the `build-ci` script are just referencing `build`. Can't we use the standard build script?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,67 @@
...
+          scope: "${{ env.SCOPE }}"
+
+      - name: Check tag suffix
+        id: check-tag-suffix
+        if: startsWith(github.ref, 'refs/tags/v') && (contains(github.ref, '-rc') || contains(github.ref, '-beta') || contains(github.ref, '-test'))
+        run: |
+          echo "::set-output name=npm-tag::next"
+
+      - name: Prepare package.json for publishing
+        run: npm run prod-config-ci
```
danielelisi, line 43: Same as for the `build-ci` script; why can't we use the normal `prod-config` script?

taqueria-protocol/package.json
```diff
@@ -4,7 +4,9 @@
   "description": "A TypeScript package which contains types that are to be shared between @taqueria/node-sdk and @taqueria/taqueria.",
   "main": "index.js",
   "scripts": {
```
mweichert, line 6: @alexzbusko Clever ;)

mweichert (commented):
Looks good. There was just one comment around the flextesa plugin waiting for other plugins to be built.

.github/workflows/main.yml
danielelisi: I spoke to @alexzbusko this needs to be tweaked to work with reusable workflow so it's okay to remove for now

danielelisi (approved):
Had a chat with @alexzbusko and we'll re-address my comments later as this PR is currently blocking other task.

Approving this PR :+1:

GImbrailo (approved):

mweichert (approved):

PR #166, opened 2022-01-29
fix(ci): build flextesa dependency for docker

GImbrailo:


GImbrailo:
@ecadlabs/sre @jevonearth  as this is after hours and needs to be merged as a fix I am going to force merge without approval from `CODEOWNERS`

PR #171, opened 2022-01-30
Convert repo to use NPM workspaces

mweichert:
All taqueria NPM packages were referring to first-party dependencies by using local file paths in the package.json files - e.g.:
`@taqueria/node-sdk": "file:../taqueria-sdk"`

However, for published NPM packages, we can't use the file protocol to refer to a package and instead have to specify the version of the published package:
`@taqueria/node-sdk": "^0.0.1"

To achieve that, we developed a `npm run prod-config` script that used sed to modify the package.json and package-lock.json files. This seemed like it was going to work, but there were more complicated modifications required to the lock file that would have been very difficult to achieve. Thus, I converted the repo to use npm workspaces which is designed to solve this problem and make monorepos easier to work with.

To do that, I had to add this to the root package.json file:
```
"workspaces": [
    "./taqueria*"
  ]
  ```
  
  With that single change, quite a lot is achieved and how you interact with the repo changes quite a bit as well:
  - You only ever run `npm install` from the root of the repo. It creates a node_modules in the root of the repo that is shared by all other packages
  - The node_modules folder contains package dependencies as normal, however, if one of the packages in this monorepo were to another package in this subrepo, there is a symlink created. The package isn't fetched from a registry.
  
As I said, using the `file:../taqueria-sdk"` notation for dependencies breaks `npm publish`, so I've adjusted all instances of those to just use a "^0.0.1" value instead. When used with NPM workspaces, everything just works as it should.

#### Other changes
- Now that we don't need to modify the package.json for CI/CI, I've removed some NPM scripts that were commonly defined: `get-sdk-version`, `get-protocol-version`,  `update-sdk-version`, `update-protocol-version`,`backup-config`, `restore-config`, `prod-config`, `build-prod`.
- In the npm-publish.yml workflow, we don't need to change the working directory to the package directory. We run all npm scripts now from the root of the repo. We target individual packages by using `npm run [script] --workspace=[dir]`. To achieve that, I define a WORKSPACE environment variable with the "dir" variable passed in to the workflow.
- I removed all package-lock.json files. There is now just one in the root.

---
```diff
- Important!
```

You should now run NPM scripts from the root of the repo. Don't cd into a sub-directory to execute an NPM script for a particular package. Instead, just specify what package you're targetting by running `npm run [script] --workspace=[dir]` - e.g.: `npm run build-ci --workspace taqueria-sdk`.

---

Overall I feel this is a much easier setup to reason about.


mweichert:
@danielelisi when you get a chance, do you mind looking into why the packaging of the vscode extension fails on Windows? 
https://github.com/ecadlabs/taqueria/runs/4999228758?check_suite_focus=true

.github/workflows/npm-publish.yml
```diff
@@ -32,35 +29,44 @@ jobs:
         with: 
           registry-url: ${{ env.NPM_REGISTRY }}
           scope: "${{ env.SCOPE }}"
-
+        env:
```
alexzbusko, line 32: @mweichert if you add an env var above the steps definition it will be made available to all steps in the workflow. So if the `GIT_WORKSPACE` env var was moved the same place as `NODE_AUTH_TOKEN` then it will be available in the entire job and you won't need to add it to every step

mweichert: Ah, I figured that there was a way. Thanks! I'll make that adjustment.

alexzbusko (approved):

Issue #53, opened 2022-01-05
`taqueria --version` should output tag or branch it was built from and the git hash


jevonearth:
### User Story
As a: taqueria cli user

I want to: run `taqueria --version`

So that I can: see which version of taqueria I have running and what git sha my taqueria build was made from 

### Acceptance Criteria
- [ ] `--version` outputs version info (long version)
- [ ] `-v` outputs version info (short version)
- [ ] Output displays the git tag that taqueria was built from (when built from a tag)
- [ ] Output displays the git branch that taqueria was built from (when built from a branch)
- [ ] Output displays the git sha that taqueria was built from

PR #173, opened 2022-01-30
Add taq version

mweichert:
Adds a --version option to the CLI which will display the branch/tag as the version #.


mweichert:
Fixes #53

PR #184, opened 2022-01-31
FIX Npm PR Version

ricklove:


ricklove:
The above version is incorrect. The actual version that exists in npm is `0.0.0-pr-184-d55e830f`

WRONG: `0.0.0-pr-184-55e830f`
CORRECT: `0.0.0-pr-184-d55e830f` <-- missing the first character of the last segment

PR #186, opened 2022-01-31
ci: Use local template reference for Github reusable workflows

danielelisi:
The Github reusable workflows introduced in #156 are tricky to work with when trying to update the template workflow. Currently we need to explicitly point to the branch we are working on in each reusable workflow instantiation. This is messy as we can easily forget to point the workflow back to `main` branch before merging a PR.

Github introduced [Reusable workflows can be referenced locally](https://github.blog/changelog/2022-01-25-github-actions-reusable-workflows-can-be-referenced-locally/ ) on Jan 25th 2022 which should solve this issue most users are experiencing issues https://github.community/t/allow-reusable-workflows-to-be-located-at-arbitrary-locations-and-be-local/212745/5?u=danielelisi

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -129,7 +129,7 @@ jobs:
   publish-flextesa-plugin-to-npm:
     needs:
       - publish-sdk-to-npm
-    uses: ecadlabs/taqueria/.github/workflows/npm-publish.yml@main
+    uses: ./.github/workflows/npm-publish.yml
     with:
       dir: taqueria-plugin-flextesa
     secrets:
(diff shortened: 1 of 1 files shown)


danielelisi:
We're experiencing the same errors as other users when using local references for reusable workflows in https://github.community/t/allow-reusable-workflows-to-be-located-at-arbitrary-locations-and-be-local/212745/5?u=danielelisi

```
Invalid workflow file : .github/workflows/main.yml#L1
The workflow is not valid. ecadlabs/taqueria/.github/workflows/npm-publish.yml@78eedbc985a777e6b5ae4a7ea9a5688bae882fcf: Expected document start parse event Unexpected type '' encountered while reading 'root'. The type 'MappingToken' was expected.
```

danielelisi:
The error in my last comment has been fixed. Using `uses: ./.github/workflows/npm-publish.yml` in reusable workflow now point to the template in the current commit. I triggered a new pipeline in this MR which completed successfully.

alexzbusko (approved):
I am very happy this is now working

GImbrailo (approved):
LGTM

PR #189, opened 2022-02-01
Attempt to fix vscode packaging

mweichert:
VSCode package requires node_modules to be included in the VXIS package, but now that we're using NPM workspaces, I believe that node_modules is in the root of the repo.

So I'm copying node_modules from the root before packaging.

In addition, the VSCode extension runs some child processes which need some adjustment to work on Windows.


danielelisi (approved):
Tested the built artifact on my machine and confirmed that it works

LGTM :+1:

Issue #206, opened 2022-02-02
[Feature]: Support multiple platforms for @taqueria


mweichert:
### Describe the feature you would like

We'd like to be able to run native docker images on platforms like linux/arm64/v8 (mac M1) without having to ask the developer to setup and install Rosetta. 

Also, I think we want to be able to run on platforms like Raspberry PI (linux/arm64).

So in total, I think we need to build images for:
- linux/amd64
- linux/arm64
- linux/arm64/v8

### Is this feature solving a problem?

Ability to run builds natively on more than just amd64.

### Please share any additional details you have?

This idea came out of a debugging session with @russellmorton, trying to figure out why some of our uses of docker command would success with the --platform argument specified in some contexts but failing in other contexts.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #207, opened 2022-02-02
feat(ci): flexesa docker image on multiple platforms

GImbrailo:
Publishing docker images for `linux/amd64` and `linux/arm64/v8`

**Testing Plan**
Please see the [taqueria-flextesa](https://github.com/ecadlabs/taqueria/pkgs/container/taqueria-flextesa) images.

image

**Note**
We found that `linux/arm64` and `linux/arm64/v8` are essentially the same platform and will work for both if only one is specified.

Ref #206

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -52,11 +52,21 @@ jobs:
       - name: Build package
         run: npm run build -w taqueria-plugin-flextesa/docker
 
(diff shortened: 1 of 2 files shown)


mweichert:
Thanks Alex!

danielelisi (approved):

mweichert (approved):

PR #222, opened 2022-02-04
added in testing steps to main workflow

alexzbusko:
This PR adds in a separate job to run the tests in the repo as part of the pipeline to make sure that PRs are getting some testing applied to them. The changes are as follows:

- Updated flextesa plugin job to run after sdk is published
- Added run tests job to make sure that dependencies are installed, binary downloaded for running tests then running all tests
- Updated unit test with changes Ivan supplied to me since it was failing locally

LARGE NOTE:
There is a change in this PR that will affect running these tests locally. I have added a README.md file to outline some of tests functionality


.github/workflows/main.yml
```diff
@@ -128,15 +128,56 @@ jobs:
...
+      - name: install dependencies for e2e tests in root and test directories
+        run: |
+          cd ..
+          curl -fsSL https://deno.land/x/install/install.sh | sh
+          npm ci
+          npm install --workspace=taqueria-plugin-ligo
+          npm install --workspace=taqueria-plugin-mock
+          
+          cd tests/
+          npm ci
```
danielelisi, line 165: You could add `./tests` to the list of npm workspaces and change use
```
npm ci --workspace taqueria-plugin-ligo
npm ci --workspace taqueria-plugin-mock
npm ci --workspace test
```

.github/workflows/main.yml
```diff
@@ -128,15 +128,56 @@ jobs:
...
+      
+      - name: Run unit tests
+        run: |
+          export PATH="/home/runner/.deno/bin:$PATH"
+          npm run test:unit
+
+      - name: Run integration tests
+        run: |
+          export PATH="$(pwd):$PATH"
```
danielelisi, line 174: Why do you need to update `PATH` here?

.github/workflows/main.yml
```diff
@@ -128,15 +128,56 @@ jobs:
...
+
+      - name: get binary and make executable from previous step for tests
+        run: |
+          wget ${{ needs.build-binaries.outputs.public-url-ubuntu }} -O taq
+          chmod +x taq
+
+      - name: install dependencies for e2e tests in root and test directories
+        run: |
+          cd ..
+          curl -fsSL https://deno.land/x/install/install.sh | sh
```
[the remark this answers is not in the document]
alexzbusko: I totally forgot that this exists despite it being used in the same file above. I will update things to install deno through the action

.github/workflows/main.yml
alexzbusko: My ignorance of workspaces is showing. I will update it to include the tests folder as a workspace

.github/workflows/main.yml
alexzbusko: @danielelisi I did not know there is an env var for path available in these runners. I have updated the starting step that gets the binary to set path for all subsequent steps

danielelisi: Sounds good. Please remove all the redundant `export PATH` on lines `174`, `179`, `184`

.github/workflows/main.yml
```diff
@@ -128,15 +128,61 @@ jobs:
...
+
+      - name: Run integration tests
+        run: |
+          export PATH="$(pwd):$PATH"
+          npm run test:integration
+
+      - name: Run e2e tests
+        run: |
+          export PATH="$(pwd):$PATH"
+          npm run test:e2e
```
danielelisi, line 185: Once you remove the default `working-directory` you'll need to run these commands from root folder with `npm run test:e2e --workspace test`.

.github/workflows/main.yml
```diff
@@ -128,15 +128,61 @@ jobs:
...
+      run:
+        working-directory: ./tests
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+
+      - uses: denoland/setup-deno@v1
```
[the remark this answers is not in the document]
alexzbusko: I'll make it specific to use the latest one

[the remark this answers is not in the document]
danielelisi: According to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-versioned-actions `v1` pins it to the latest major version. So it's better to keep it as you originally wrote it so that we don't need to update it manually

danielelisi (approved):
LGTM 👍

tests/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@taqueria/test",
+  "name": "@taqueria/tests",
```
mweichert, line 2: Can we add a `"private":  true` to the package.json please? This will ensure that this package never accidentally gets published. Also, helps communicate that to others looking here.
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private

mweichert (changes requested):
Thanks for taking this on Alex. Looks good. I suggested some very small and petty adjustments.

tests/package.json
alexzbusko: I have updated this to be private

tests/README.md
```diff
@@ -0,0 +1,24 @@
+# Taqueria Tests
+## Usage:
+Since the taqueria project is currently making use of npm workspaces all activity is generally going to take place in the root (taqueria) directory. Commands will not function normally if run from the tests directory and the following commands will be what you can use to install all dependencies and run the tests for taqueria:
```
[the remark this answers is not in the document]
alexzbusko: Not at all, I have updated it

mweichert (approved):
You da man! Thanks!

PR #274, opened 2022-02-09
Fixes #273, LIGO compilation failures on Apple Silicon

### LINKED ISSUE #273: [Bug]: LIGO plugin - On Apple Silicon, displays "There was a compilation error" although compilation succeeds
On an Apple M1, if you use the LIGO plugin to compile a smart contract you will always get an error message indicating that the compilation failed. However, that's usually a false positive (see #254).

As such, we need to refactor the LIGO plugin to not emit a docker warning if possible.

mweichert:
Sets the DOCKER_DEFAULT_PLATFORM environment variable when running the _ligolang/ligo` docker image. Necessary to avoid docker warnings until the LIGO team is able to add multi-platform builds for their image.

Fixes #273


mweichert:
FYI, passes all LIGO end-to-end tests.

russellmorton (approved):

PR #299, opened 2022-02-10
267 run unit tests on commit

alexzbusko:
This is a PR to add a commit hook that will run the unit tests whenever a user commits some changes. It assumes the following:

- User has run npm install
- User has installed deno as part of the setup for unit testing

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npm install -w tests && npm run test:unit -w tests
+npm run test:unit -w tests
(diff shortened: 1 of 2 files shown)


mweichert (approved):

MichalLytek:
@mweichert @alexzbusko 
I find running tests on commit as an antipattern.
Tests should be run on every commit but on CI, not on a local machine.
Developer should run the test locally whenever he wants.

Running tests on commit makes the whole commit experience super slugish.
We now have only 1 unit tests, I can't imagine waiting 30 seconds or a few minutes on every commit.
My workflow is to commit small changes frequently and this prevent me from doing this.

We should only do quick checks like ensuring the codebase is well formatted (with the help of dPrint PR #695).
I used to think that pre-commit hook is a good place for linting the code but it also makes the commit slower and irritating for a developer.
If we do the `-fix` version, it may commit some unwanted changes, like removing comments or changing variable names.
If it warns and stop commit, it's super annoying.

I've opened a new PR to remove this pre-commit hook - #719.

PR #301, opened 2022-02-10
Fixes #295, installing a plugin in the VSCode extension

### LINKED ISSUE #295: [Bug]: Receive error message "There was a problem running taq." when adding a plugin
### What happened?

When installing a plugin via the command pallette to a a Taqified project, and error pop up is shown saying "There was a problem running taq."

The plugin still installs correctly. For example, I added the LIGO plugin, received the error but then was able to compile a test contract.

The plugin should install without showing the user an error.

Using VS Code and the Taqueria extenstion v0.0.8:

1. From the command palette run `Taqueria: Initialize Project`
2. Select the project directory
3. In the terminal run `npm init -y` on the project directory
4. From the command palette run `Taqueria: Install Plugin`
5. Select `@taqueria/plugin-ligo` from the dropdown
6. Error will show in the bottom right:
7.  From the command palette run `Taqueria: Compile (LIGO)`
8. Contract will compile to  artifacts folder

### Relevant log output

_No response_

### Screenshots and images

Screen Shot 2022-02-09 at 3 10 39 PM

### Code of Conduct

mweichert:
## What the problem was

I discovered that the reason why plugins were failing in VSCode is when a version of v16 wasn't installed on the user's computer.

The SDK uses a third-party package called stack-trace. In its package.json, a requirement of nodejs v16 is hardcoded. If you have v17, when `taq install` runs `npm install` under-the-hood, and NPM will write a warning on stderr.

Taqueria interprets this as an error (as it probably should).

## How I solved it

I forked the stack-trace package and adjusted it to require nodejs >= 16, as the package runs in v17 just fine. I then installed this fork in the SDK. I created a [pull request](https://github.com/felixge/node-stack-trace/pull/39) for the maintainer of this package that we can revert back to once the pull request is accepted and merged in.

I also made an adjustment to the vscode extension so that if a package emits a EBADENGINE warning that we prompt the user to install Node v16.

As the vscode extension requests the latest published package package from npmjs.com, we won't be able to verify that this PR works. I might create a separate issue for that.


mweichert:
@alexzbusko as we can't really test this till we generate new packages, I thought I'd ask if you'd mind giving this PR a quick review? Thanks!

taqueria-plugin-flextesa/package.json
```diff
@@ -36,13 +36,13 @@
...
     "fast-glob": "^3.2.7",
-    "promise-retry": "^2.0.1",
-    "@taqueria/node-sdk": "^0.0.5"
+    "promise-retry": "^2.0.1"
   },
   "devDependencies": {
     "@types/promise-retry": "^1.1.3",
     "parcel": "^2.0.1",
-    "typescript": "^4.5.2"
+    "typescript": "4.5.4"
```
alexzbusko, line 46: Do we want to specifically lock in this version of typescript? Why are we getting rid of the ^?

mweichert: Hey Alex. Yep, that was deliberate. I mentioned this in a commit message but should have also mentioned it in the PR description - 4.5.5 has a bug: https://github.com/microsoft/TypeScript/issues/32408

Aside from that, as TypeScript is a core dependency, I think it does indeed make sense to lock down, as it's shared among many workspaces in the project. Without doing so, its possible that one workspace might use 4.5.6 and another use 4.5.7 (once those versions come out). For some dependencies, that probably doesn't matter but TypeScript seems more significant.

alexzbusko (approved):

PR #329, opened 2022-02-17
CI Test

ricklove:
- Testing CI
- Fixed pre-commit command

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,7 +64,7 @@ jobs:
         env:
           TAQ_VERSION: "{{ github.ref_name }}"
         run: |
-          DENO=./deno deno compile --output taq --target $DENO_TARGET --allow-run --allow-write --allow-read --allow-env index.ts --quickstart "`cat quickstart.md`" --setBuild ${{ github.ref_name }} --setVersion ${{ github.ref_name }}
+          DENO_DIR=./deno deno compile --output taq --target $DENO_TARGET --allow-run --allow-write --allow-read --allow-env index.ts --quickstart "`cat quickstart.md`" --setBuild ${{ github.ref_name }} --setVersion ${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         env:
           TAQ_VERSION: "{{ github.ref_name }}"
         run: |
-          deno compile --output taq --target $DENO_TARGET --allow-run --allow-write --allow-read --allow-env index.ts --quickstart "`cat quickstart.md`" --setBuild ${{ github.ref_name }} --setVersion ${{ github.ref_name }}
+          DENO_DIR=./deno deno compile --output taq --target $DENO_TARGET --allow-run --allow-write --allow-read --allow-env index.ts --quickstart "`cat quickstart.md`" --setBuild ${{ github.ref_name }} --setVersion ${{ github.ref_name }}
```
GImbrailo, line 67: I tested without the using the `DENO_DIR` variable. Is there a reason why we want to use this variable?

.github/workflows/main.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Setup Deno
         uses: denoland/setup-deno@v1
         with:
-          deno-version: "^1.0.0"
+          deno-version: "1.18.2"
```
GImbrailo, line 57: There is another section in this file where we need to change the version on line 151

.github/workflows/main.yml
GImbrailo: @mweichert

.github/workflows/main.yml
GImbrailo: @mweichert

.github/workflows/main.yml
GImbrailo: Sounds good

.github/workflows/main.yml
mweichert: Updated.

PR #331, opened 2022-02-17
Fixes #330, type definitions incorrect for Fluture

### LINKED ISSUE #330: [Bug]: Type definitions incorrect
At the early stages, I must have been testing something and defined the Future type like so:

```
export type Future<L,R> = unknown
```

Whereas, the type should have been:
```
export type Future<L,R> = FutureInstance<L, R>
```

I suspect that I was testing Skypack for importing the fluture library. The module generated from Skypack is missing all type definitions for Fluture.

Unfortunately, this was accidentally committed some time ago.

We need to change the type definition to the latter example above. In addition, we need to replace the skypack import with the ES module that Fluture provides for Deno:
```
import type {FutureInstance} from 'https://cdn.jsdelivr.net/gh/fluture-js/Fluture@14.0.0/dist/module.js'
```

mweichert:
This PR does the following:
- Adjusts the type definition for a Future to `export type Future<L,R> = FutureInstance<L, R>`
- Stops importing Fluture from Skypack and instead uses: 'https://cdn.jsdelivr.net/gh/fluture-js/Fluture@14.0.0/dist/module.js'
- Fixes all incorrect type definitions that were missed due to an incorrect type definition for Fluture previously


taqueria-config.ts
```diff
@@ -83,35 +83,39 @@ export const getConfigPath = (projectDir: SanitizedAbsPath, configDir: Sanitized
...
         readJsonFile(configPath),
-        chainRej ((err:unknown) => {
+        chainRej (err => {
             if (!create) return reject(err)
             else {
                 return pipe(
-                    writeJsonFile(configPath) (defaultConfig),
-                    chain ((configPath: string) => readJsonFile<Config>(configPath))
+                    writeJsonFile<Config>(configPath) (defaultConfig),
+                    chain<TaqError, string, Config> ((configPath: string) => readJsonFile(configPath))
```
ricklove, line 95: Is this better?

```
chain((configPath) => readJsonFile<Config>(configPath))
```

npm.ts
```diff
@@ -58,19 +64,19 @@ const addToPluginList = (pluginName: NpmPluginName, config: ConfigArgs) => pipe(
 
 export const installPlugin = (configDir: SanitizedPath, projectDir: SanitizedAbsPath, i18n: i18n, plugin: string) => pipe(
     requireNPM(projectDir, i18n),
-    chain (() => exec('npm install -D <%= it.plugin %>', {plugin}, projectDir)),
-    chain (() => getConfig(projectDir, configDir, i18n, false)),
-    chain ((config: ConfigArgs) => {
+    chain(_ => exec('npm install -D <%= it.plugin %>', {plugin}, projectDir)),
+    chain<TaqError, number, ConfigArgs>(_ => getConfig(projectDir, configDir, i18n, false)),
```
ricklove, line 68: Same, seems to work without the types here as well:
```
    chain(_ => getConfig(projectDir, configDir, i18n, false)),
```

npm.ts
```diff
@@ -22,14 +26,16 @@ export const getPluginName = (input:string): NpmPluginName => {
     return (retval as NpmPluginName)
 }
 
-export const requireNPM = (projectDir: SanitizedAbsPath, i18n: i18n) => pipe(
-    readJsonFile(projectDir.join("package.json").value),
-    chainRej (() => reject({kind: 'E_NPM_INIT', msg: i18n.__("npmInitRequired"), context: projectDir}))
+export const requireNPM = (projectDir: SanitizedAbsPath, i18n: i18n) : Future<TaqError, Manifest> => pipe(
+    readJsonFile<Manifest>(projectDir.join("package.json").value),
+    chainRej<TaqError, TaqError, Manifest>(previous => reject({kind: 'E_NPM_INIT', msg: i18n.__("npmInitRequired"), context: projectDir, previous})),
```
ricklove, line 31: Seems to work without the types expressed:
```
chainRej(previous => reject({kind: 'E_NPM_INIT', msg: i18n.__("npmInitRequired"), context: projectDir, previous}))
```

ricklove (approved):
Looks good, just a few considerations

taqueria-utils/taqueria-utils.ts
```diff
@@ -32,33 +33,26 @@ const mkdirFuture = (path: string): Future<TaqError, void> => attemptP(() => Den
...
+            .catch((previous: Error) => Promise.reject({
+                kind: "E_READ", msg: `Could not read ${path}`, previous
+            }))
+    })
 
 export const readJsonFile = <T>(path: string) => pipe(
     readTextFile(path),
-    chain (decodeJson),
-    map ((result: unknown) => (result as T))
+    chain<TaqError, string, T> (decodeJson)
```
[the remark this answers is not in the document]
mweichert: Yep, I like that better too.

taqueria-config.ts
mweichert: Good find.

PR #336, opened 2022-02-18
Allow string or Sanitized util params

ricklove:
This increases the complexity of the util function implementations, but improves their usability especially in a pipe.


hu3man:
Looks like this PR is waiting on your feedback @mweichert . Looks feature complete

PR #434, opened 2022-03-03
Fixes #408, adds @taqueria/flextesa-manager workspace and NPM package

### LINKED ISSUE #408: Move flextesa startFlextesa package
The @taqueria/plugin-flextesa package contains a docker folder, which is also an NPM package and contains a script called startFlextesa.ts, which is used by docker image we generate for managing flextesa instances.

The startFlextesa.ts script uses @taqueria/node-sdk.

This repo is configured to use workspaces, but nested workspaces doesn't seem to be supported. So we can't have one workspace for taqueria-plugin-flextesa and then have a nested workspace of taqueria-plugin-flextesa/docker. 

Currently, taqueria-plugin-flextesa/docker is using the version of the @taqueria/node-sdk from npmjs.com rather than the version from the repo, which makes development difficult.

I propose that we move taqueria-plugin-flextesa/docker to taqueria-flextesa-manager so that we can manage it like any other workspace.

This requires alterations to the CI/CD process as well as the flextesa plugin.

mweichert:
Fixes #408 by adding a separate workspace at the root of the repo called `@taqueria/flextesa-manager`.


alexzbusko (approved):
LGTM

danielelisi (approved):

PR #436, opened 2022-03-03
added matrix build for tests in CI

alexzbusko:
This PR covers the following:
- Includes the macos-latest test runner for running all of our tests on a mac host to better reflect our users environments
- Installs colima(https://github.com/abiosoft/colima) as a container runtime to get docker on mac for the different tests that require it. It is based on lima(https://github.com/lima-vm/lima) which is also something to keep an eye
- Moves the notifications for e2e test failures into a separate job that will only trigger if the run-tests job fails since the action will not run on a mac host
- Updated the e2e test timeout to 3 minutes for some tests taking longer running on a mac host
- Workflow level concurrency based on branch name to cancel jobs for the same branch if a new one is scheduled before a currently running one is complete


hu3man:
@alexzbusko , what's the status of this?

alexzbusko:
> 

@hu3man I initially made this as a POC for running our tests in a mac context but the docker setup on a mac runner takes a really long time (ubuntu tests <5 minutes while the mac ones are ~16 minutes) so I just let it be. There is no native support for docker on mac and one thing I discussed with Gino in the past was a self-hosted runner that we could use for this purpose but it has fallen by the wayside in favour of other things

hu3man:
Thanks for the context @alexzbusko . GIven the issues now surfacing with the Docker/Flextesa images on Mac, is this something we should revisit @mweichert ?

GImbrailo (approved):
@alexzbusko LGTM. I see the CF build job is failing and should work again if you merge the latest changes in the `main` branch

hu3man (approved):
This looks good to me @alexzbusko! Runs as expected and I don't see any issues in the code.  Takes quite a bit of time to run on mac-os so short-circuiting in-progress runs on the same branch is a smart move 👍

jchenche (approved):

sinapsist (approved):
LGTM

PR #438, opened 2022-03-04
add archetype plugin in ts

guillaumeduhamel:

PR #461, opened 2022-03-09
rafactor(ci): Only run-tests job for PRs from a fork

GImbrailo:
We are encountering errors in our CICD when workflows are triggered from forked repositories e.g #438 . These workflows fail because of some of our jobs publish assets and require write permissions. This PR adds conditions to the workflows to only run the `run-tests` job when a PR is made from a fork. 

### Testing plan
image
*Only run-tests job for fork PRs*

https://github.com/ecadlabs/taqueria/actions/runs/1959955142

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -181,6 +181,13 @@ jobs:
         with:
           deno-version: "1.18.2"
 
+      - name: download the binary from the build-binaries job and make it executable
+        run: |
+          wget ${{ needs.build-binaries.outputs.public-url-ubuntu }} -O taq
+          chmod +x taq
+          echo "$(pwd)" >> $GITHUB_PATH
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+
       - name: build binary and make executable
         env:
           TAQ_VERSION: ${{ github.ref_name }}
@@ -191,6 +198,7 @@ jobs:
           [[ $(./taq init -p ./test_project) == "Project taq'ified!" ]] 
           chmod +x taq
           echo "$(pwd)" >> $GITHUB_PATH
+        if: ${{ github.event.pull_request.head.repo.fork }}
 
       - name: install dependencies
         run: npm ci
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -174,9 +174,14 @@ jobs:
         with:
           deno-version: "1.18.2"
 
-      - name: get binary and make executable from previous step for tests
+      - name: build binary and make executable
+        env:
+          TAQ_VERSION: ${{ github.ref_name }}
+          DENO_DIR: "./deno"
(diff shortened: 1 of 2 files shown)


GImbrailo:
@mweichert @alexzbusko could you please provide feedback on the proposed changes in the PR description?

.github/workflows/main.yml
```diff
@@ -174,9 +181,14 @@ jobs:
         with:
           deno-version: "1.18.2"
 
-      - name: get binary and make executable from previous step for tests
+      - name: build binary and make executable
+        env:
+          TAQ_VERSION: ${{ github.ref_name }}
+          DENO_DIR: "./deno"
```
alexzbusko, line 187: Is DENO_DIR even needed? I don't see where it is used

.github/workflows/main.yml
```diff
@@ -174,9 +181,14 @@ jobs:
...
         run: |
-          wget ${{ needs.build-binaries.outputs.public-url-ubuntu }} -O taq
+          COMMIT=`git rev-parse --short "$GITHUB_SHA"`
+          deno compile --output taq --allow-run --allow-write --allow-read --allow-env index.ts --quickstart "`cat quickstart.md`" --setBuild "$TAQ_VERSION/$COMMIT" --setVersion "$TAQ_VERSION"
```
alexzbusko, line 190: I think it's worthwhile to split this out and make this local compile only happen when coming from a forked repo and download the file from GCP otherwise. That way for internal PRs we still download the binary for specific build targets and forked PRs will have this generic binary built for them.

I mostly want to leave the door open for my other PR that will have the tests run on each binary that we produce (linux and mac)

alexzbusko (changes requested):
I left two comments. The main part I would like to discuss is what binary the tests will run on when coming from a forked repo vs our repo

.github/workflows/main.yml
GImbrailo: Yeah it is used by when the `deno` command executes

BGMurphy (approved):
Other than the changes Alex mentioned, LGTM!

.github/workflows/main.yml
GImbrailo: Sure thing. I added a another step and a condition that would use one or the other based on the PR coming from a fork or not. Because we would need the `build-binaries` job to complete first even if it is skipped, I had to add another two conditions to the `run-tests` job.

.github/workflows/main.yml
alexzbusko: Alright cool

.github/workflows/main.yml
alexzbusko: Awesome, thank you so much Gino

alexzbusko (approved):
Looks great

Issue #388, opened 2022-02-25
Re-enable test to verify that taqueria taquito plugin can deploy one contract using deploy {contractName} command


sinapsist:
After this issue is fixed (https://github.com/ecadlabs/taqueria/issues/387) - need to re-enable the test and verify that it works correctly.
- [x] #387

mweichert:
Fixed in #465

Issue #373, opened 2022-02-23
Investigate test "Verify that taqueria ligo plugin will display proper message if user tries to compile contract that does not exist" and enable it after stdout fixed 


sinapsist:
As an SDET I want to investigate why taq compile on file that does not exist is not placing error to the stdout (more details can be found here https://github.com/ecadlabs/taqueria/issues/372). After the issue is fixed or a workaround found enable or fix and enable the test.

Issue #260, opened 2022-02-09
Flextesa plugin - refactored to use stdout, stderr, and response objects


mweichert:
### Describe the feature you would like

Once #257 is done, the Flextesa plugin will need to be refactored to output to stdout, stderr, and use the SDK to generate response objects.

### Is this feature solving a problem?

Please see #257 for more details.

### Please share any additional details you have?

Will use #258 as a reference.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #258, opened 2022-02-09
LIGO plugin - refactored to use stdout, stderr, and response objects


mweichert:
### Describe the feature you would like

Once #257 is done, the LIGO plugin will need to be refactored to output to stdout, stderr, and use the SDK to generate response objects.

### Is this feature solving a problem?

See #257 

### Please share any additional details you have?

Details about the problem your feature solves

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #465, opened 2022-03-11
Fixes #257, protocol refactor for improved stdout/stderr handling

### LINKED ISSUE #257: Refactor handling of stdout/stderr and response object for protocol

Plugins are currently expected to return a JSON serialized message on stdout or stderr like the following:
```
{
  status: 'success',
  stdout: '',
  stderr: '',
  render: 'table'
}
```

This implementation at the protocol layer was an experiment and not something we were committed to. Indeed, there are problems with this approach which will be outlined later in this issue.

Instead, the plugin should be able to write to stdout and stderr and have that relayed by the Taqueria CLI. The plugin should also be able to output a stream that doesn't have to consumed in a memory buffer on the Taqueria-side or plugin-side.

Lastly, we need a means of outputting a structured response object from the plugin to Taqueria. The transport should support multiple encodings, but for now, its fine if Taqueria only has the ability to work with JSON encodings.

The structured response object could also be an exception or TaqError, and therefore will need to be able to accommodate multiple response types.

To test the refactoring, within the scope of this issue we should also refactor the type generation plugin.

The current implementation has the following problems:
- It isn't a pleasant developer experience for plugin authors. It's natural for a developer to expect that they can just log output using `console.log` or `console.error`.
- It imposes an unnecessary restriction for a plugin developer to wrap their output.
- It makes plugin authoring and debugging harder. A developer can't output something to console.log for debugging purposes.
- It requires buffering stdout and stderr. The plugin can't provide streaming output

Some thoughts about how to consume the output on the Deno/Taqueria side:
https://stackoverflow.com/questions/66861364/stream-stdout-from-subprocess

I would emphasize that in the implementation and refactoring effort we keep the following in mind:

> How do we keep the protocol itself (transport between plugins and Taqueria) simple and concise? By keeping the protocol itself small and with fewer restrictions and opinions, it becomes easier for plugin authors to understand. It also reduces friction for developers looking to develop an SDK for another language or platform, such as a Python SDK.

### Code of Conduct

mweichert:
Fixes #257, #258, #259, #260, #261, #262, and #479.

Highlights:
- Plugin tasks now have an encoding field, that defaults to "none", but accepts either "json" or "application/json" as a value. When "none", stdout and stderr isn't buffered, and output directly to the stdout/stderr of the Taqueria process. This is known as "passthru".

- Plugin functionality has segmented from the cli.ts file into plugin.ts, and should have 100% unit test coverage. This file exports a function called "inject", which receives dependencies to inject as input, and outputs a collection of functions with their dependencies injected. Examples of dependencies are stdout and stderr, which allows us to test output effectively.

- Houston, for plugins.ts, I've added a comment header above each function to provide some documentation. We should talk about what tool(s) we want to use for documentation generation. Ideally what's there can serve as a starting point for you to work with.

- Helper methods for emitting output and errors from a plugin task have been added to the Taqueria Node SDK. Note, the SDK still has no unit test coverage.

- Each plugin has been refactored to use an appropriate encoding for its tasks. 

- We used to pass '--logPluginCalls' when debugging why a plugin. To keep terminology consistent with rest of the codebase, I've renamed this to '--logPluginRequests'. Also, if you use this flag, which outputs the command used to invoke a plugin, you don't have to manually copy/paste it - it will exist in your clipboard automatically.

- You should never see a JSON object emitted as an error - only error messages.


mweichert:
Fixes #373 and #372.

mweichert:
Fixes #388 and #387

PR #500, opened 2022-03-18
added in step to notify channel when e2e tests fail

alexzbusko:
Added in a step to notify the e2e notifications channel in slack when a job fails


hu3man (approved):
LGTM!

PR #518, opened 2022-03-24
feat(ci): add CodeQL workflows

GImbrailo:
This PR adds the CodeQL workflow to enable code scanning on a scheduled basis as well as for pull requests

### Testing plan

image
*Example of issues found by CodeQL*

image
*Example of CodeQL finding no issues after fix*

https://github.com/ecadlabs/taqueria/security/code-scanning?query=pr%3A518+tool%3ACodeQL

image
*Searchable history of branch issues*

### DIFF
--- a/taqueria-protocol/taqueria-protocol-types.ts
+++ b/taqueria-protocol/taqueria-protocol-types.ts
@@ -82,7 +82,7 @@ const commandType: unique symbol = Symbol()
 export class Command extends StringLike {
     [commandType]: void
     static create(value: string) : Command | undefined {
-        if (value.match(/^([A-Za-z-_ ]+ ?)((\[[^\]\[\n]*\] ?)|(<[^<>\n]+>) ?)*$/)) {
+        if (value.match(/^([A-Za-z-_ ]+ ?)((\[.+\] ?)|(\<.+\>) ?)*$/)) {
             return new Command(value)
         }
         return undefined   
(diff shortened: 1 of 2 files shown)


BGMurphy (approved):
LGTM

taqueria-protocol/taqueria-protocol-types.ts
```diff
@@ -82,7 +82,7 @@ const commandType: unique symbol = Symbol()
 export class Command extends StringLike {
     [commandType]: void
     static create(value: string) : Command | undefined {
-        if (value.match(/^([A-Za-z-_ ]+ ?)((\[.+\] ?)|(\<.+\>) ?)*$/)) {
+        if (value.match(/^([A-Za-z-_ ]+ ?)((\[[^\]\[\n]*\] ?)|(<[^<>\n]+>) ?)*$/)) {
```
danielelisi, line 85: is this an unrelated change?

GImbrailo: Yes, this change was added to show what happens when an issue that CodeQL finds is fixed. Removing this change will cause the workflow to fail which I think is acceptable. There are also some other options to dismiss the CodeQL findings that are shown here:

image

GImbrailo: I removed the change

danielelisi (approved):

PR #520, opened 2022-03-26
Contract type plugin fix 02

ricklove:
- Display console.error and warn
- Handle `pair` with more than 2 items


hu3man (approved):
This LGTM

PR #570, opened 2022-04-08
Closes #561 add missing typechecking to sdk and flextesa manager plugin

### LINKED ISSUE #561: [Bug]: Explicit type inconsistency
### What happened?

After reviewing the source code, we've found a style inconsistency. Specifically, given the type definition below, the handler field of UnvalidatedTask isn't making use of `TaskHandler`

`export type TaskHandler = "proxy" | string | string[]`

Variables of type `"proxy" | string | string[]` should specify `TaskHandler` instead.

Look at `taqueria/taqueria-protocol/taqueria-protocol-types.ts` and `taqueria/taqueria-sdk/types.ts`

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jchenche:
It's initially meant to fix a code style, but later discovered that type checking was missing in the sdk and flextesa manager plugin workspaces.

Closes #561

### DIFF
--- a/taqueria-flextesa-manager/package.json
+++ b/taqueria-flextesa-manager/package.json
@@ -4,7 +4,7 @@
   "description": "Provides an interface and abstraction over a flextesa mininet",
   "main": "index.js",
   "scripts": {
-    "build-package": "npx esbuild --platform=node --bundle index.ts --outdir=.",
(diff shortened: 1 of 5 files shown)


hu3man (approved):
Good catch on the type checking Jimmy! LGTM and builds as expected locally but it looks like there is a failure at the comment step on the GH pipeline

Approving so it doesn't hold up release, but let's ensure the pipeline is fixed and all green before merging

mweichert (approved):

PR #575, opened 2022-04-11
refactor(docs): build documentation from a single source

GImbrailo:
The changes introduce a script `parse-readme-template.ts` that makes use of Eta and templates to generate two sets of docs, one `README.md` that makes use of GitHub flavoured markdown, and a `_taqueria-${plugin_name}.mdx` document for the website. A list of plugins is defined in the script and only those plugins' docs will be parsed. You must ensure that there is a `_readme.eta` template in the plugin directory with the document contents. The `eta` and `ts-node` packages have been added to the dev dependencies. 

To manually run this script you can run `npm run parse-readme-template` in the project root directory which will basically run `ts-node ./parse-readme-template.ts`.

A husky pre-commit hook has been added that will automatically run `ts-node ./parse-readme-template.ts`, and add any changes to any `.md` or `.mdx` to the commit. This way you do not have to run the script manually when committing changes to the plugin documentation.


website/docs/plugins/plugin-typescript-generator.mdx
```diff
@@ -127,3 +127,128 @@ const contract = await Tezos.contract.at<TestContract>(`tz123`);
 const storage = await contract.storage();
 ```
```
hu3man, line 129: Should this not use the import like the other files?

hu3man (approved):
Works great @GImbrailo!

Approving now as it's feature complete and functional. Just have a few discussion points 

We should document how this templated build system works so new contributors can understand it. If you can provide some high level steps/points, I can write them into a doc

Topics we should cover:
- What steps need to be taken to add a new plugin with docs to the project
- Substitution/template syntax use and rules
- The execution flow that occurs during the build process

Aside from that, we just need to rebase the branch and migrate the Tezos Client docs. Let me know if you need a hand with that!

mweichert (approved):
Hey @GImbrailo

Looks good to me, although should we remove the n-alex directories?

website/docs/plugins/plugin-typescript-generator.mdx
GImbrailo: Thanks for pointing that out. I pushed a fix and added some more detail to the PR description

hu3man:
Hey @GImbrailo, was trying to merge this branch with main but the package-lock.json merge conflict is pretty gnarly. What's the best way to handle this?

PR #579, opened 2022-04-12
Fixes #578, e2e tests not running locally

### LINKED ISSUE #578: Fix e2e tests not running locally
There is a problem with running e2e tests locally using NPM v7 or later that results in each test case failing with:

```
    File did not exists and was not created during the timeout.

      69 |         const timer = setTimeout(function () {
      70 |             watcher.close();
    > 71 |             reject(new Error('File did not exists and was not created during the timeout.'));
         |                    ^
      72 |         }, timeout);
      73 |
      74 |         fs.access(filePath, fs.constants.R_OK, function (err) {

      at Timeout._onTimeout (e2e/utils/utils.ts:71:20)
```

The problem occurs because `npm init -y` behaves different in NPM v7. If it detects that workspaces are enabled, and you try to run `npm init -y` within a defined workspace, such as _tests_, then it will assume that the NPM command should be targeting the _tests_ workspace.

mweichert:
NOTE: This shouldn't resolve the scaffolding e2e tests.

See https://github.com/ecadlabs/taqueria/issues/578

I've worked around the issue by running `npm init -y` outside of the workspace, and then moving the test project directory back into the workspace. This is necessary as our test cases make assumptions that the directory is relative to the tests/e2e directory.

### DIFF
--- a/npm.ts
+++ b/npm.ts
@@ -41,7 +41,10 @@ export const requireNPM = (projectDir: SanitizedAbsPath, i18n: i18n) : Future<Ta
 )
 
 export const getPluginPackageJson = (pluginNameOrPath: string, projectDir: SanitizedAbsPath) => pipe(
-    readJsonFile(SanitizedAbsPath.create(pluginNameOrPath, projectDir).join('package.json').value),
+    /^\//.test(pluginNameOrPath) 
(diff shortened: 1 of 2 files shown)


sinapsist (approved):
LGTM

PR #582, opened 2022-04-15
Add Prettier to project codebase

MichalLytek:
Currently, all the codebase is randomly formatted. Some files are using 2 spaces indentation, some are using 4 spaces. Some are using `'` for string, some are using `"`. Semicolon is used in single line of code, while the one below does not have it. Promises and other method chains are indented chaotically.

Prettier standardize the formatting, forcing only and consistent option. This makes the code much more readable and (when using format on save) lift the responsibility of maintaining proper formatting out of the developer's head.

This PR adds prettier config, along with a npm tasks to format and check formatting.
Also all the codebase has been formatted, so no ghost changes will appear in next PRs.

We can discuss and change some prettier config settings to better much our developers needs.


MichalLytek:
I have read and agree to the CLA

PR #583, opened 2022-04-18
Fixes #512, uses bulk operations API for Taquito originations

### LINKED ISSUE #512: [Bug]: Deploying multiple contracts can fail
### What happened?

If someone executes `taq originate`, we will deploy all contracts from the artifacts directory. We do this by using Taquito's contract.originate() method. We currently do this sequentially, which can cause problems, and in particular on testnets.

Lets say that there are two contracts to deploy, A and B. The operation for originating contract A is sent to the node with a particular counter, such as 1000. The operation for deploying contract B is sent but without waiting for confirmation that the first operation has been accepted in a block. Due to low gas, volume, or otherwise, the first operation is still in the mempool, and therefore the second operation has the same counter.

This causes the following error to be emitted:
```
HttpResponse: Http error response: (500) [{"kind":"temporary","id":"failure","msg":"Error while applying operation ooZofWLwKpCJWZ1Avfw4yCGS1UXaHyKjUWKSZZziEygEL9aU9K1:\nError:\n  Counter 3339520 already used for contract tz1UjBsGr8Tkogbz8fcySP6x77S5TKBvs99t (expected 3339521)\n"}]
```

We should use the batch API. No errors should be thrown.

Try to deploy multiple contracts to a testnet. High likelihood that it will fail.

### Relevant log output

```shell
{
  errCode: 'E_ORIGINATE',
  errMsg: "Could not originate hello-tacos-one.tz on the network 'hangzhounet'",
  previous: HttpResponse: Http error response: (500) [{"kind":"temporary","id":"failure","msg":"Error while applying operation ooZofWLwKpCJWZ1Avfw4yCGS1UXaHyKjUWKSZZziEygEL9aU9K1:\nError:\n  Counter 3339520 already used for contract tz1UjBsGr8Tkogbz8fcySP6x77S5TKBvs99t (expected 3339521)\n"}]
  
      at XMLHttpRequest.request.onload (/Users/mweichert/Projects/taqueria/node_modules/@taquito/http-utils/dist/taquito-http-utils.umd.js:430:30)
      at XMLHttpRequestEventTarget.dispatchEvent (/Users/mweichert/Projects/taqueria/node_modules/xhr2-cookies/dist/xml-http-request-event-target.js:34:22)
      at XMLHttpRequest._dispatchProgress (/Users/mweichert/Projects/taqueria/node_modules/xhr2-cookies/dist/xml-http-request.js:358:14)
      at XMLHttpRequest._onHttpResponseEnd (/Users/mweichert/Projects/taqueria/node_modules/xhr2-cookies/dist/xml-http-request.js:319:14)
      at IncomingMessage.<anonymous> (/Users/mweichert/Projects/taqueria/node_modules/xhr2-cookies/dist/xml-http-request.js:289:61)
      at IncomingMessage.emit (node:events:377:35)
      at endReadableNT (node:internal/streams/readable:1312:12)
      at processTicksAndRejections (node:internal/process/task_queues:83:21) {
    status: 500,
    statusText: 'Internal Server Error',
    body: '[{"kind":"temporary","id":"failure","msg":"Error while applying operation ooZofWLwKpCJWZ1Avfw4yCGS1UXaHyKjUWKSZZziEygEL9aU9K1:\\nError:\\n  Counter 3339520 already used for contract tz1UjBsGr8Tkogbz8fcySP6x77S5TKBvs99t (expected 3339521)\\n"}]\n',
    url: 'https://hangzhounet.api.tez.ie/injection/operation'
  }
}
```
```


### Screenshots and images

Screenshot and images

### Code of Conduct

### PR DESCRIPTION
**Pull request by mweichert**

### CODE REVIEW

File: taqueria-plugin-taquito/originate.ts [hunk 1]
```diff
@@ -17,185 +24,167 @@ interface OriginationResult {
...
+                    if (network.faucet) {
+                        const result = (async () => {
+                            const tezos = new TezosToolkit(network.rpcUrl as string)
+                            await importKey(tezos, network.faucet.email, network.faucet.password, network.faucet.mnemonic.join(' '), network.faucet.activation_code)
+                            return await createBatch(parsedArgs, tezos, networkName)
+                        })()
+
+                        return [...retval, result]
+                    }
+                    else sendErr(`Network ${networkName} requires a valid faucet`)
```
hu3man, line 127: For best ux, could we point the user to where they fix it with something like:
```
Network ${networkName} requires a valid faucet to be configured in `config.json`
```?

hu3man (approved):
Looks good @mweichert! 

The implementation for using batched network/sandbox req's  lgtm and will add my approval however I don't have much context about the test implementation so I'll leave that for others to review/approve

taqueria-plugin-taquito/originate.ts
mweichert: Good idea. I've made that adjustment.

alexzbusko:
@mweichert it looks good to me. One thing I don't understand is the tests seem to be taking a bit longer than they were before which makes me wonder what is up. I was also testing out reducing the time to let the contracts propagate in the network to 25s. I would say that is something to do here since it's a small change but if you don't I'm sure it will happen in the future

alexzbusko (approved):

PR #614, opened 2022-04-22
made installs for plugins default to local

alexzbusko:
Changed the commands spec to use local installation for plugins to reduce total time that spec takes to execute. I will make another PR in the future to create a spec to test the installation from the npm registry to make sure that is covered


hu3man (approved):
LGTM,

One question, will the e2e tests still use the NPM install for plugins during the release?

mweichert (approved):
LGTM!

PR #618, opened 2022-04-25
Fixes #557 typechecker and simulate task

### LINKED ISSUE #557: Michelson typecheck task as part of Tezos Client Plugin
As a smart contract developer, I want to type check Michelson code using Taqueria so that I can validate the smart contract before deploying it.

jchenche:
Implementation and e2e tests for `typecheck` and `simulate` tasks of the **flextesa** plugin


cli.ts
```diff
@@ -413,6 +413,7 @@ const addTask = (cliConfig: CLIConfig, config: ConfigArgs, env: EnvVars, parsedA
             )
         },
         handler: (inputArgs: Record<string, unknown>) => {
+            // console.log("======JCC\n", inputArgs.storage, "\n======")
```
alexzbusko, line 416: I think this can be removed if it is no longer being used

tests/e2e/data/typechecker-simulator-contents.ts
```diff
@@ -0,0 +1,55 @@
+export const oneRowTable = `
+┌────────────────┬────────┐
+│ Contract       │ Result │
+├────────────────┼────────┤
+│ hello-tacos.tz │ Valid  │
+└────────────────┴────────┘
+`.trimStart()
+
+export const oneRowTableSimulateResult = `
+┌────────────────┬────────────────────┐
```
alexzbusko, line 10: Is this how we want the output to look?

jchenche: Yep this is the best solution after discussing with @mweichert

tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -235,4 +237,231 @@ describe("E2E Testing for taqueria flextesa plugin",  () => {
...
+    beforeAll(async () => {
+        dockerName = "local"
+        await generateTestProject(taqueriaProjectPath, ["flextesa"]);
+        await exec(`taq start sandbox ${dockerName}`, {cwd: `./${taqueriaProjectPath}`})
+    })
+
+    test('Verify that taqueria typechecker task can typecheck one contract under contracts folder', async () => {
+        try {
+            // 1. Copy contract from data folder to taqueria project folder
+            execSync(`cp e2e/data/hello-tacos.tz ${taqueriaProjectPath}/contracts`);
```
alexzbusko, line 253: We have had problems with mixing sync and async functions in the past. Can you try to use async functions when you can?

tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -235,4 +237,231 @@ describe("E2E Testing for taqueria flextesa plugin",  () => {
...
+
+            // 2. Run taq typecheck
+            const {stdout, stderr} = await exec(`taq typecheck ${dockerName}`, {cwd: `./${taqueriaProjectPath}`});
+
+            // 3. Verify that it's well-typed and contains no errors
+            expect(stdout).toContain("Valid");
```
alexzbusko, line 259: Your assertion here is checking that valid is in the output but you have made a file for the output contents. Can you make the assertion check that the `oneRowTable` is what is output?

taqueria-plugin-flextesa/index.ts
```diff
@@ -49,6 +49,57 @@ Plugin.create(_i18n => ({
             ],
             encoding: "json"
         }),
+        Task.create({
+            task: "typecheck",
+            command: "typecheck <sandboxName> [sourceFiles...]",
```
hu3man, line 54: Coming from the UX standpoint here:
1. Can you provide some use cases for when a user would be using the sandboxName optional parameter?
2. Why is [sourceFiles] a required parameter? Can it not default to type checking all files in the configured artifacts directory? (This would be a more intuitive ux)
3. Why does the handler not start the default sandbox if it's not running rather than returning an error message to the user?

hu3man, line 54: Typo: Remove the trailing period `.` at the end of the error messages

Bug:
It appears that the sandboxName parameter is required in use, this should be optional or removed altogether

Output:
```
taqueria/333 on  557-typechecker-task-JCC 
❯ taq tc taco.tz
There is no sandbox configuration with the name taco.tz.
```

tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -235,4 +237,231 @@ describe("E2E Testing for taqueria flextesa plugin",  () => {
...
+    });
+
+    test('Verify that taqueria typechecker task will display proper message if user tries to typecheck contract that does not exist', async () => {
+        try {
+            // 1. Run taq typecheck ${contractName} for contract that does not exist
+            const {stdout, stderr} = await exec(`taq typecheck ${dockerName} test.tz`, {cwd: `./${taqueriaProjectPath}`})
+
+            // 2. Verify that output includes a table and an error message
+            expect(stdout).toBe(contents.nonExistent)
+            expect(stderr).toContain("Does not exist");
```
alexzbusko, line 328: So when it comes to the output of error conditions do we want to output both the error and the standard output? Is this case the table and a message are printed giving the user double of the information. It looks like the following:
```
test.tz: Does not exist

┌──────────┬────────────────┐
│        Contract     │             Result                   │
├──────────┼────────────────┤
│        test.tz         │         Does not exist          │
└──────────┴────────────────┘
```

tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -235,4 +237,231 @@ describe("E2E Testing for taqueria flextesa plugin",  () => {
...
+    test("Verify that taqueria typechecker task emits error and yet displays table if contract is ill-typed", async () => {
+        try {
+            // 1. Copy contract from data folder to taqueria project folder
+            execSync(`cp e2e/data/hello-tacos-ill-typed.tz ${taqueriaProjectPath}/contracts`)
+
+            // 2. Run taq typecheck hello-tacos-ill-typed.tz
+            const {stdout, stderr} = await exec(`taq typecheck ${dockerName} hello-tacos-ill-typed.tz`, {cwd: `./${taqueriaProjectPath}`})
+
+            // 3. Verify that output includes a table and an error message
+            expect(stdout).toBe(contents.typeError)
```
alexzbusko, line 344: Same as the above for the double output

taqueria-plugin-flextesa/index.ts
```diff
@@ -49,6 +49,57 @@ Plugin.create(_i18n => ({
...
+                    description: "The name of the sandbox to use"
+                }),
+                PositionalArg.create({
+                    placeholder: "sourceFiles",
+                    description: "The names of the Michelson contracts you wish to typecheck, separated by space"
+                }),
+            ],
+            encoding: "json"
+        }),
+        Task.create({
```
hu3man, line 71: Currently, the output from a failed command does not show the user what short flags/arguments are missing

The current output is: 
```
taqueria/333 on  557-typechecker-task-JCC 
❯ taq simulate local taco.tz
taq simulate <sandboxName> <sourceFile>

Run Michelson contracts as a simulation

Positionals:
  sandboxName  The name of the sandbox to use                         [required]
  sourceFile   The name of the Michelson contract you wish to simulate, separate
               d by space                                             [required]

Options:
      --version     Show version number                                [boolean]
      --build       Display build information about the current version[boolean]
  -p, --projectDir  Path to your project directory               [default: "./"]
      --configDir   Config directory (default ./.taq)        [default: "./.taq"]
  -e, --env         Specify an environment configuration
      --help        Show help                                          [boolean]
  -s, --storage     The initial storage used to run the script        [required]
  -i, --input       The input used to run the script                  [required]

Missing required arguments: storage, input
```

The output for the command should include the short flag options as part of the structure of the command, and details below so a user immediately understand what is required when the task fails. Given that there are 4 required arguments for the command, we should not expect the user to remember them.

Something along the lines of::
```
❯ taq simulate 
taq simulate sandboxName sourceFile -s '{initial storage: 1}' -i 'format of input here'

Run Michelson contracts as a simulation

Positionals:
  sandboxName  The name of the sandbox to use                         [required]
  sourceFile   The name of the Michelson contract you wish to simulate, separate
               d by space                                             [required]

Required Parameters:
Initial storage: -s        Describe format for initial storage here
Input: -i                        Desicrbe format for input here
```

taqueria-plugin-flextesa/index.ts
```diff
@@ -49,6 +49,57 @@ Plugin.create(_i18n => ({
...
+            task: "simulate",
+            command: "simulate <sandboxName> <sourceFile>",
+            aliases: ["sim"],
+            description: "Run Michelson contracts as a simulation",
+            options: [
+                Option.create({
```
hu3man, line 77: Can we be more explicit about the expected input format for both options? (ie string? JSON? quotes? etc)

hu3man (commented):
Looks good for the most part!

Biggest question I have is why it seems to be using the `/contracts` dir rather than the configured /artifacts` dir for Michelson source files?

Also have a few ux related concerns

hu3man (commented):
Looks good for the most part!

Biggest question I have is why it seems to be using the `/contracts` dir rather than the configured /artifacts` dir for Michelson source files?

Also have a few ux related concerns

cli.ts
jchenche: Will reflect in the next commit

tests/e2e/taqueria-plugin-flextesa.spec.ts
jchenche: You mean use `await exec` instead of `execSync`?

tests/e2e/taqueria-plugin-flextesa.spec.ts
jchenche: Will come in the next commit

tests/e2e/taqueria-plugin-flextesa.spec.ts
hu3man: I think this is due to the same issue of outputting both stdout and stderr but wanted to add this output for context: 
```
taqueria/333 on  557-typechecker-task-JCC 
❯ taq simulate local taco.tz --storage '' -i ''  
 
Error:
  Erroneous command line argument 6 ().
  empty expression

Usage:
  tezos-client [global options] command [command options]
  tezos-client --help (for global options)
  tezos-client [global options] command --help (for command options)
  tezos-client --version (for version information)

To browse the documentation:
  tezos-client [global options] man (for a list of commands)
  tezos-client [global options] man -v 3 (for the full manual)

Global options (must come before the command):
  -d --base-dir <path>: client data directory (absent: TEZOS_CLIENT_DIR env)
  -c --config-file <path>: configuration file
  -t --timings: show RPC request times
  --chain <hash|tag>: chain on which to apply contextual commands (possible tags are 'main' and 'test')
  -b --block <hash|tag>: block on which to apply contextual commands (possible tags are 'head' and 'genesis')
  -w --wait <none|<int>>: how many confirmation blocks before to consider an operation as included
  -p --protocol <hash>: use commands of a specific protocol
  -l --log-requests: log all requests to the node
  -A --addr <IP addr|host>: [DEPRECATED: use --endpoint instead] IP address of the node
  -P --port <number>: [DEPRECATED: use --endpoint instead] RPC port of the node
  -S --tls: [DEPRECATED: use --endpoint instead] use TLS to connect to node.
  -m --media-type <json, binary, any or default>: Sets the "media-type" value for the "accept" header for RPC requests to the node. The media accept header indicates to the node which format of data serialisation is supported. Use the value "json" for serialisation to the JSON format.
  -E --endpoint <uri>: HTTP(S) endpoint of the node RPC interface; e.g. 'http://localhost:8732'
  -s --sources <path>: path to JSON file containing sources for --mode light. Example file content: {"min_agreement": 1.0, "uris": ["http://localhost:8732", "https://localhost:8733"]}
  -R --remote-signer <uri>: URI of the remote signer
  -f --password-filename <filename>: path to the password filename
  -M --mode <client|light|mockup|proxy>: how to interact with the node

┌──────────┬─────────┐
│ Contract │ Result  │
├──────────┼─────────┤
│ taco.tz  │ Invalid │
└──────────┴─────────┘
```

jchenche: I think it's not a big deal. But I can change the output inside the table from "Does not exist" to "N/A"? Would that be a better?

tests/e2e/taqueria-plugin-flextesa.spec.ts
jchenche: This is different from the above. The output looks like this, which looks fine:

```
jimmychenchen@Jimmys-MBP taco % taq tc local hello-tacos-ill-typed.tz
 
Ill typed contract:
  01: { parameter nat ;
  02:   storage string ;
  03:   code { UNPAIR ;
  04:          SWAP ;
  05:          DUP ;
  06:          DUG 2 ;
  07:          SWAP ;
  08:          DUP ;
  09:          DUG 2 ;
  10:          COMPARE ;
  11:          GT ;
  12:          IF { DROP 2 ; PUSH string "NOT_ENOUGH_TACOS" ; FAILWITH }
  13:             { SWAP ; SUB ; ABS ; NIL operation ; PAIR } } }
  14: 
  15: 
At line 10 characters 9 to 16, wrong stack type for instruction COMPARE:
  [ nat @parameter : string @storage : ... ].
Bad stack item 1.
Type nat is not compatible with type string.
Fatal error:
  ill-typed script

┌──────────────────────────┬─────────┐
│ Contract                 │ Result  │
├──────────────────────────┼─────────┤
│ hello-tacos-ill-typed.tz │ Invalid │
└──────────────────────────┴─────────┘

```

taqueria-plugin-flextesa/index.ts
jchenche: Angle brackets means it's required and square bracket means it's optional. So sandbox is required, whereas source files are not. In the absence of source files, it will typecheck all source files.

I wanted to let the users start the sandbox. The error message will prompt them to start one. It also makes the implementation simpler for now.

tests/e2e/taqueria-plugin-flextesa.spec.ts
hu3man: It should not output any of the tezos-client specific information as the user is unable to use any of that error messaging through the abstraction

taqueria-plugin-flextesa/index.ts
hu3man: My bad on the brackets, however, in the command output it does show both as being required right now: 
```
Positionals:
  sandboxName  The name of the sandbox to use                         [required]
  sourceFile   The name of the Michelson contract you wish to simulate, separate
               d by space                                             [required]
```

jchenche: The sandbox is actually required (see my response to another comment above, with substring: "Coming from the UX standpoint here:" ).

hu3man: What is the argument for having a user specify the sandbox?

Is there a circumstance where the user may want to use a sandbox other than local? I can see the need for protocol related type-checking perhaps, but coupling the sandbox this tightly to the type-checker seems like an anti-pattern to me and makes the ux much less intuitive

taqueria-plugin-flextesa/index.ts
jchenche: I think this UI improvement might be outside of the scope of this project. It works like this in every other plugin. e.g. `taq compile --help`

taqueria-plugin-flextesa/index.ts
hu3man: That I understand, but why can't the command use the `default` sandbox specified in `config.json`?

taqueria-plugin-flextesa/index.ts
hu3man: That is another case where it's optional (can be run with or without `--help`)

If it's an architectural limitation, could we not also use positional arguments for initial storage and input?

taqueria-plugin-flextesa/index.ts
jchenche: Due to a limitation in parsing commands by the Taqueria CLI, I made sandbox mandatory. Otherwise, we'd have optional sandbox and optional source files, and we can't parse this.

taqueria-plugin-flextesa/index.ts
jchenche: Ok will reflect in the next commit

tests/e2e/taqueria-plugin-flextesa.spec.ts
jchenche: Hmm I will see if I can trim it to just:

```
Error:
  Erroneous command line argument 6 ().
  empty expression
```

taqueria-plugin-flextesa/index.ts
jchenche: I believe this is for simulation. For typechecking, sourceFile is optional.

The user should be able to just use any sandbox and not have to specify one. But I made it mandatory for now due to command parsing limitations in the CLI.

taqueria-plugin-flextesa/index.ts
jchenche: I don't think it makes sense to use propositional arguments for storage and input. They are, by nature, options. Propositionals are CLI or plugin related, but Options are task-related, as in they are meant to be passed to some underlying process that the tasks invokes

jchenche: The bottom of the error output, `Missing required arguments: storage, input`, prompts the user to specify them.

tests/e2e/taqueria-plugin-flextesa.spec.ts
mweichert: @jchenche I think displaying the reason why a contract isn't valid makes sense. However, if there's an error due to incorrect usage of the tezos-client, then we shouldn't output the message from tezos-client but try to prevent the end-developer from incorrect invocation.

tests/e2e/taqueria-plugin-flextesa.spec.ts
mweichert: 👍

taqueria-plugin-flextesa/index.ts
mweichert: 👍

mweichert (changes requested):
Awesome job, @jchenche! Looks great to me. One question is whether you think we'd be best to break up proxy.ts into multiple files. Maybe it makes sense to have the following structure:
- proxy/lib.ts (common functions used by the files below)
- proxy/index.ts
- proxy/manage.ts
- proxy/typecheck.ts
- proxy/simulate.ts

What do you think?

tests/e2e/taqueria-plugin-flextesa.spec.ts
jchenche: Will reflect in the next commit

taqueria-plugin-flextesa/index.ts
hu3man: Can we at least make the error output at the bottom indicate the structure of the command, or what parameters are missing? (taq simulate sandboxName sourceFiles --storage {} --input {})?

jchenche:
> Awesome job, @jchenche! Looks great to me. One question is whether you think we'd be best to break up proxy.ts into multiple files. Maybe it makes sense to have the following structure:
> 
> * proxy/lib.ts (common functions used by the files below)
> * proxy/index.ts
> * proxy/manage.ts
> * proxy/typecheck.ts
> * proxy/simulate.ts
> 
> What do you think?

I thought about breaking it down to something similar. Should it be part of this PR or should I create an issue to refactor this?

taqueria-plugin-flextesa/index.ts
hu3man: Ok, will solve with added documentation for now

tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -235,4 +237,231 @@ describe("E2E Testing for taqueria flextesa plugin",  () => {
...
+            // 3. Verify that output includes a table and an error message
+            expect(stdout).toBe(contents.runtimeError)
+            expect(stderr).toContain("unexpected string, only an int can be used here");
+
+        } catch(error) {
+            throw new Error (`error: ${error}`);
+        }
+    })
+
+    test("Verify that taqueria simulator task emits parameter type error (supplying list instead of nat)", async () => {
```
alexzbusko, line 417: I don't think this is strictly necessary since we have a test that tries a string and the error states that the input can only be an integer. I think only one test is needed since this isn't testing anything new

tests/e2e/data/help-contents.ts
```diff
@@ -216,6 +216,10 @@ Commands:
                                                                  [aliases: stop]
   taq list accounts <sandboxName>           List the balances of all sandbox acc
                                             ounts
+  taq typecheck <sandboxName> [sourceFiles  Typecheck Michelson contracts
+  ...]                                                             [aliases: tc]
+  taq simulate <sandboxName> <sourceFile>   Run Michelson contracts as a simulat
+                                            ion                   [aliases: sim]
```
alexzbusko, line 222: There should also be test for the individual commands for the help menu. I think there should be two more entries here similar to helpContentsFlextesaPluginTypeCheck and helpContentsFlextesaPluginSimulate

tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -235,4 +237,231 @@ describe("E2E Testing for taqueria flextesa plugin",  () => {
...
+            // 3. Verify that output includes a table and an error message
+            expect(stdout).toBe(contents.runtimeError)
+            expect(stderr).toContain("NOT_ENOUGH_TACOS");
+
+        } catch(error) {
+            throw new Error (`error: ${error}`);
+        }
+    })
+
+    test("Verify that taqueria simulator task emits parameter type error (supplying string instead of nat)", async () => {
```
alexzbusko, line 400: The title says nat but the error is specifying int, is the actual restriction to natural numbers?

tests/e2e/taqueria-plugin-flextesa.spec.ts
alexzbusko: Alternatively there are tests to specify a bad type value for input but not for storage. I think that one should also be covered and this test could be rewritten to cover that

jchenche: I was trying to test that the format to specify a list is valid. e.g. '{1;2;3}'

tests/e2e/data/help-contents.ts
jchenche: We will be moving all to a new plugin after discussion with @mweichert and @hu3man

tests/e2e/taqueria-plugin-flextesa.spec.ts
jchenche: That's an error emitted by Tezos-client's simulation command. But yes, the actual restriction is to natural integer (no negative numbers)

tests/e2e/taqueria-plugin-flextesa.spec.ts
alexzbusko: Alright cool, my dyslexia hit and I didn't notice the work sequence was different from string in the previous test

taqueria-plugin-tezos-client/client.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+
+export const client = <T>(parsedArgs: Opts) : LikeAPromise<void, Failure<T>> => {
+    switch (parsedArgs.task) {
+        case 'typecheck':
+            return typecheckTask(parsedArgs)
+        case 'simulate':
+            return simulateTask(parsedArgs)
+        default:
+            return sendAsyncErr(`${parsedArgs.task} is not an understood task by the Tezos-client plugin`,)
```
hu3man, line 252: This messaging should be improved

At a minimum:
```
${parsedArgs.task} is not a valid Taqueria task
```

Ideally something actionable like:
```
${parsedArgs.task} is not a valid task. Tezos-client plugin tasks are: /n
taq typecheck
taq simulate
```

hu3man (commented):
Looks and works great @jchenche!

Left a few comments and am running into the issue I mentioned on slack. Adding those details here too:

Running into an error when running `taq simulate` using storage defined in `config.json`. I get the following:
```
❯ taq simulate tacos.tz '2'              
 
storage.replace is not a function
┌──────────┬────────────────────────────┐
│ Contract │ Result                     │
├──────────┼────────────────────────────┤
│ tacos.tz │ Bad input or storage value │
└──────────┴────────────────────────────┘
```

When running it and passing storage to the `taq simulate` command, it works fine:
❯ taq simulate tacos.tz '2' --storage '5'
┌──────────┬────────────────────┐
│ Contract │ Result             │
├──────────┼────────────────────┤
│ tacos.tz │ storage            │
│          │   3                │
│          │ emitted operations │
│          │                    │
│          │ big_map diff       │
│          │                    │
│          │                    │
└──────────┴────────────────────┘
```

File: taqueria-plugin-tezos-client/client.ts [hunk 14]
```diff
@@ -0,0 +1,256 @@
...
+    }
+    return sendAsyncErr(`There is no sandbox configuration with the name ${parsedArgs.sandboxName}.`)
+}
+
+//////////// Simulate task ////////////
+
+const getStorageFromConfig = (opts: Opts, sourceFile: string) : string|undefined => {
+    const config = opts.config
+    const defaultEnv = config.environment.default
+    const storages = config.environment[defaultEnv].storage
```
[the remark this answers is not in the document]
jchenche: For now, it only supports the default environment. But it wouldn't be hard to extend.

tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -235,4 +237,236 @@ describe("E2E Testing for taqueria flextesa plugin",  () => {
...
-});
\ No newline at end of file
+});
+
+// TODO: to be moved into a different spec file when the tezos-client plugin no longer depends on the flextesa plugin.
+// If I move it now, testing won't work.
+// Starting docker in e2e/auto-test-flextesa-plugin and then e2e/auto-test-tezos-client-plugin won't work
+// because Jest parallelizes test runs across spec files it seems.
+// The volumes end up being different, and so the tests below will complain that files don't exist.
+describe("E2E Testing for taqueria typechecker and simulator tasks of the tezos-client plugin", () => {
```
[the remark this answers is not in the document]
jchenche: Due to Jest parallelization (more detail in the code comment above the test suite), I am leaving it in flextesa.spec for now. It would be easy to move it to a new spec file in the future; just need to cut the entire `describe()` function. But I don't have the expertise make it work today.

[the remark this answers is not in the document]
jchenche: I have a solution that doesn't require some sort of special Jest knowledge. But that requires a quite significant amount of changes.

The solution is to have an extra parameter for the container name (`docker run --name <the name>`) when starting one. Right now it's "hardcoded" (in a sense) to be the same as the sandbox name.

taqueria-plugin-tezos-client/client.ts
jchenche: Taking a second look, I think this should never happen. That would be a mismatch between `index.ts` and this case statement.

taqueria-plugin-tezos-client/client.ts
```diff
@@ -0,0 +1,256 @@
...
+    return sendAsyncErr(`Please specify a contract. E.g. taq simulate local <contractName> ...`)
+}
+
+const simulateTask = async <T>(parsedArgs: Opts) : LikeAPromise<void, Failure<T>> => {
+    const sandbox = getSandbox(parsedArgs)
+    if (sandbox) {
+        if (doesUseFlextesa(sandbox)) {
+            return await isSandboxRunning(sandbox.name.value)
+            ? simulate(parsedArgs, sandbox).then(sendJsonRes)
+            : sendAsyncErr(`The ${sandbox.name} sandbox is not running.`)
```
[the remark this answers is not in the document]
jchenche: I think this is only useful to users who just started using Taqueria. After a while, it'd be quite verbose and there'd be extra syncing to be done in case we change the UI for starting a sandbox?

I don't have a strong preference.

taqueria-plugin-tezos-client/client.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    // If the string contains escaped double quotes, escape them further
+    storage = storage.replace(/\\"/g, '\\\\\\"')
+    input = input.replace(/\\"/g, '\\\\\\"')
+
+    // If the string contains leading and trailing double quotes, escape them, otherwise docker exec will complain
+    storage = storage.match(/^".*"$/) ? "\\" + storage.slice(0, -1) + "\\\"" : storage
+    input = input.match(/^".*"$/) ? "\\" + input.slice(0, -1) + "\\\"" : input
+
+    // TODO: maybe validate storage and input value before passing it to tezos-client to prevent tezos-client menu being displayed?
```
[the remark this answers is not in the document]
jchenche: I think we can either trim everything down starting from "Usage" or do data validation as we go (@mweichert suggests the latter if I recall). Since it's pretty isolated, I suppose it could be a separate "enhancement issue"?

taqueria-plugin-tezos-client/client.ts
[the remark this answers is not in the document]
hu3man: Agreed, and the more I think about it as well, the real solution is to have intuitive software that can detect and recover from these types of easily handled errors and that's a future issue

taqueria-plugin-tezos-client/client.ts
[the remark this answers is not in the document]
hu3man: Given the low chance of someone actually encountering this, a quick and dirty solution is fine here for now.

If we defer the validation work, let's make sure to log the in a separate ticket

taqueria-plugin-tezos-client/client.ts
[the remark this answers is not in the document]
hu3man: With the current tasks implemented, the only use case I can think of would be to run the `taq typecheck` task against 2 different protocols using the `-e` cli flag

But, since we can currently handle this right in the task with the `--sandbox` option/flag, it's not actually a product gap so you can decide whether to implement now or defer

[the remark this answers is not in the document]
jchenche: Since any sandbox will work for now, I will defer until we need it

taqueria-plugin-tezos-client/client.ts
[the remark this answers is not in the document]
jchenche: hmm on second thought, I feel like trimming seems to be a much easier solution and it covers all cases (have `tezos-client` do all the validation so that we are not re-inventing the wheel).

tests/e2e/taqueria-plugin-flextesa.spec.ts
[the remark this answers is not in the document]
hu3man: I like that approach @jchenche 👍

On a related, but different note:

Right now there is an issue where we can only launch a single sandbox at a time (even when different ports are specified:
```
❯ taq start sandbox local
Error: Command failed: sh -c "docker run --name local --rm --detach --platform linux/arm64/v8 -p 20000:20000 -p 19229:9229 -v /Users/houstonkrohman/code/taqueria-scaffold-demo:/project -w /app ghcr.io/ecadlabs/taqueria-flextesa:dev-557-typechecker-task-JCC-200406f node index.js --sandbox local"
docker: Error response from daemon: driver failed programming external connectivity on endpoint local (2dc5d41c650b90ac7b05d80557407a395bd0b0a2da35da9ffdfb96adb3d034c2): Bind for 0.0.0.0:19229 failed: port is already allocated.

    at ChildProcess.exithandler (node:child_process:399:12)
    at ChildProcess.emit (node:events:520:28)
    at maybeClose (node:internal/child_process:1090:16)
    at Socket.<anonymous> (node:internal/child_process:449:11)
    at Socket.emit (node:events:520:28)
    at Pipe.<anonymous> (node:net:687:12) {
  killed: false,
  code: 125,
  signal: null,
  cmd: 'sh -c "docker run --name local --rm --detach --platform linux/arm64/v8 -p 20000:20000 -p 19229:9229 -v /Users/houstonkrohman/code/taqueria-scaffold-demo:/project -w /app ghcr.io/ecadlabs/taqueria-flextesa:dev-557-typechecker-task-JCC-200406f node index.js --sandbox local"'
}
```
first sandbox is running on port 2020, tried starting local on port 2000...

Is this hardcoding related to that?

taqueria-plugin-tezos-client/client.ts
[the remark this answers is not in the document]
hu3man: I think trimming it is the best solution for now

tests/e2e/taqueria-plugin-flextesa.spec.ts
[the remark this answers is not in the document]
jchenche: hmm not sure about this one.

taqueria-plugin-tezos-client/client.ts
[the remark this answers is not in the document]
jchenche: Thoughts? @mweichert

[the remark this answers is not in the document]
jchenche: I implemented trimming and it works like wonders! I think we should stick with this. It's also more maintainable because we only have to change a single line in case `tezos-client` change their menu to removing "Usage:" (it's the pattern I look for when I trim).

taqueria-plugin-tezos-client/client.ts
```diff
@@ -0,0 +1,263 @@
+import {Sandbox as theSandbox, execCmd, getArch, sendAsyncErr, sendErr, sendRes, sendAsyncRes, sendJsonRes, sendAsyncJsonRes} from '@taqueria/node-sdk'
+import type {SanitizedArgs, Config, Attributes, SandboxConfig, Failure, LikeAPromise, Sandbox, AccountDetails, StdIO, PluginResponse} from "@taqueria/node-sdk/types"
+import type {ExecException} from 'child_process'
+import retry from 'async-retry'
+import {join} from 'path'
+import glob from 'fast-glob'
+
+type Opts = SanitizedArgs & {sandboxName?: string}
+
+const isSandboxRunning = (sandboxName: string) =>
```
mweichert, line 10: @jchenche there looks to be quite a bit of copy/paste from the @taqueria/plugin-flextesa plugin. 

I think it would be best to add the @taqueria/plugin-flextesa as a dependency to the package.json file and then just import the functions you need into scope.

taqueria-plugin-tezos-client/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    /* Projects */
+    // "incremental": true,                              /* Enable incremental compilation */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "ES6",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
```
mweichert, line 14: Do you mind bumping the target to ES2021 please?

mweichert (changes requested):
Excellent work! I did request a change with regards to some of the duplication between the flextesa and tezos-client plugins.

taqueria-plugin-tezos-client/tsconfig.json
jchenche: For sure! Coming in the next commit

taqueria-plugin-tezos-client/client.ts
jchenche: After discussion with @mweichert. We are leaving them as is. Reason being that there's only 2 being copied over and they are tiny.

hu3man (approved):
I'm happy with the functionality of the plugin / tasks at this point

Giving my approval now with the assumption that the unresolved conversations/changes will be made before merging

taqueria-plugin-tezos-client/README.md
```diff
@@ -1,13 +1,14 @@
...
 This is a plugin developed for Taqueria built on NodeJS using the Taqueria Node SDK and distributed via NPM
 
-The SmartPy plugin exposes a compile task in Taqueria which can target one, or all the SmartPy contracts in the contracts folder and compile them to Michelson code output to the artifacts folder
+The Tezos Client plugin exposes the typecheck and simulate tasks. The typecheck task will type check one or more Michelson contracts found in the /artifacts subfolder of the project root. The simulate task will dry run a Michelson contract found in the same abovementioned subfolder.
 
 ## Tasks
 
-The SmartPy plugin provides the following tasks to Taqueria:
-- `compile`
+The Tezos Client plugin provides the following tasks to Taqueria:
```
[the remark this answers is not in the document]
jchenche: Ok, will come in the next commit

mweichert (approved):

hu3man (approved):
Awesome work @jchenche!

I'll follow up with you later today regarding the docs for the website

.github/workflows/main.yml
```diff
@@ -173,6 +173,16 @@ jobs:
     secrets:
       NPM_TOKEN: ${{ secrets.NPMJS_PAT }}
 
+  publish-tezos-client-plugin-to-npm:
```
hu3man, line 176: @GImbrailo , it's this workflow here that's in the SRE wheelhouse and needing approval

GImbrailo (approved):
@jchenche LGTM

PR #662, opened 2022-05-02
Turn LIGO Plugin from JS to TS

jchenche:


taqueria-plugin-ligo/compile.ts
```diff
@@ -1,17 +1,21 @@
-const {execCmd, getArch, sendAsyncErr, sendJsonRes, sendErr} = require('@taqueria/node-sdk')
-const {extname, basename, join} = require('path')
-const glob = require('fast-glob')
+import {execCmd, getArch, sendAsyncErr, sendJsonRes, sendErr} from '@taqueria/node-sdk'
+import {SanitizedArgs, PluginResponse, Failure, LikeAPromise} from "@taqueria/node-sdk/types";
+import {extname, basename, join} from 'path'
+// @ts-ignore interop issue. Maybe find a different library later
```
mweichert, line 4: @jchenche what is the ts-ignore for?

taqueria-plugin-ligo/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    /* Projects */
+    // "incremental": true,                              /* Enable incremental compilation */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "ES6",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
```
mweichert, line 14: I believe most other packages are targeting ES2021. Do you mind making that adjustment please?

mweichert (approved):
Looks good! I did add a couple of comments/questions to address before merging, but I'm happy to see this done. Good work Jimmy!

hu3man (approved):
LGTM aside from the es2021 target version

Approving now so gtm isn't held up

taqueria-plugin-ligo/tsconfig.json
jchenche: will do!

taqueria-plugin-ligo/compile.ts
jchenche: Without it, it will give me the error: `can only be default-imported using the 'esModuleInterop' flag`

We've seen this before and solved it using `@ts-ignore` due to some 3rd-party library type issue, if I recall correctly.

mweichert: Mind trying if `import * as glob from 'fast-glob'` works please?

jchenche: Hmm it works though I don't know why the editor is complaining (same version of TS). Can we ignore the complain for VS Code?

Screen Shot 2022-05-02 at 15 14 15

hu3man:
Nice work @jchenche!

taqueria-plugin-ligo/compile.ts
jchenche: However, `import glob = require('fast-glob')` works, makes both VS Code and `tsc` happy. Thoughts @mweichert?

PR #704, opened 2022-05-09
chore: add environment variable option to compile operation

GImbrailo:
This change introduces a condition in the `ligo-plugin` compile function. The condition checks to see if `PROJECT_DIR` has been set. If so, the compile function will use that variable. This change does not take away form the current functionality

```node
    if (process.env.PROJECT_DIR && process.env.PROJECT_DIR?.length > 0) {
        projectDir = process.env.PROJECT_DIR
    }
```


taqueria-plugin-ligo/compile.ts
```diff
@@ -15,9 +15,12 @@ const getInputFilename = (opts: Opts) => (sourceFile: string) => {
 }
 
 const getCompileCommand = (opts: Opts) => (sourceFile: string) => {
-    const {projectDir} = opts
+    let {projectDir} = opts
```
mweichert, line 18: Hey @GImbrailo, I'd suggest one change:
```js
const projectDir = process.env.PROJECT_DIR ??  opts.projectDir

if (!projectDir) throw `No project directory provided`
```

Later, once my 620-operations branch is merged in, we can rewrite as:
```js
const projectDir = SanitizedAbsPath.create(process.env.PROJECT_DIR ?? opts.projectDir)
```

mweichert (approved):

taqueria-plugin-ligo/compile.ts
GImbrailo: Thanks @mweichert I implemented your suggestion and it is working as expected.

PR #695, opened 2022-05-06
As a Taqueria dev, I want my code to be automatically formatted so that the codebase is more consistent and adheres to the project style guide at all times

MichalLytek:
Superseded #582 

Depends on #719

PR #719, opened 2022-05-13
As a Taqueria dev, I want my code to be automatically tested only on pushing branch to origin, so that the codebase on GitHub is healthy but it doesn't force me to run tests on all my local commits

MichalLytek:
~I've opened this PR as I find running tests on commit as an antipattern.
Tests should be run on every commit but on CI, not on a local machine.
Developers should run the test locally whenever they want.~

~Running tests on commit makes the whole commit experience super sluggish.
We now have only 1 unit tests, I can't imagine waiting 30 seconds or a few minutes on every commit.
My workflow is to commit small changes frequently, and this prevents me from doing this.~

~We should only do quick checks like ensuring the codebase is well formatted (with the help of dPrint PR #695).
I used to think that pre-commit hook is a good place for linting the code, but it also makes the commit slower and irritating for a developer.
If we do the `-fix` version, it may commit some unwanted changes, like removing comments or changing variable names.
If it warns and stop commit, it's super annoying.~

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,8 +1,6 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npm run test:unit -w tests
-
 ## Process documentation files and add modified files to the git commit
 echo "Processing documentation files..."
 npx ts-node ./parse-readme-template.ts
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npm run test:unit -w tests


alexzbusko (approved):

mweichert:
Hi @MichalLytek,

We had a chance with everyone else on the team to review this issue and I think what we'd like to try is the following:
- We'll move the test automation to run at the pre-push hook instead of commit. This isn't necessarily the end-decision, but we'd like to try that and see how it works for everyone
- After that change has been implemented, we can try using `HUSKY=0 git push` when you know you don't want/need the tests to be run locally when pushing.

Hopefully that's a reasonable compromise for now, and we can revisit in a couple of weeks to see how everyone perceives the change and if any other adjustments are required.

Thanks,
Mike

MichalLytek:
Given my years of experience with linters, formatters, test runners and git hooks, this never ends well.
The only battle tested solution with git hooks is formatting the codebase (which is now extremely fast with dprint) to make sure even with other IDE, the commited code follows our styleguide.

I personally hate all the forced linters and test suites either on commit or push, especially when they can fail and broke the pipeline.
It's not an issue for me, because I can alias the shell command and always call `git` with `--no-verify`, or use the `HUSKY=0`.

Of course we can give it a try and see. But then for sure we need to get the feedback from developers if that really helps them or just irritates them as much as me 😉

MichalLytek:
@mweichert Changes applied, now it moves tests to the push phase 😉

mweichert (approved):

PR #729, opened 2022-05-17
702 fix tezos client empty storage jcc

### LINKED ISSUE #702: [Bug]: Tezos client plugin storage or input value of 0 is treated as empty by the code and also, yargs messes with hex values
### What happened?

Tezos client storage or input value of 0 is treated as empty by the code. Refer to screenshot below (`opts.storage` is `0`), so `if (opts.storage) {} {}` will execute the false branch, hitting the error shown at the debug line.

While fixing this. Also fix the issue that `--storage '0x00'` in the taq CLI will turn into `0` by the time the plugin receives it, as seen in the screenshot. We need to preserve `0x00`. Turns out we need to preserve all hex values because `yargs` messes with them.

0 should not be treated as empty. In other words, the if statement should execute the true branch.

Note `--storage '0x00'`

```shell
taq simulate timelock.tz 'Pair 0x8ac392f7a5a5e1befaadc891c6c0ba85baafa9d6d897feccd4c7f6a7aefef2b1bb9ba58f8594a8f9e3cddba3ebbbec81c9fb8e82cbe48db2f09ce1e5e1ed889ef68ae4e0efecd58cb4aada89e1f699e1bca19da9b4d6d5d9b2fbf3e8d783f3c8908c808fb7cfbd9ae9f3bde4c0b5edf9aecaa292cb968b82eb8896d8e9eda28cfa9febe3cd8aa4f1bad3df9adbadc8b6a382c797c294f0ec83e0bfbb819bcfe5a1fd98cafd8cb3a9eccfccf2ea98ceef9085b5f98fe1edeb84bbb58cc3fda0f98dfb85b7b5ae9ac4e0d0eee3e0cdb9cffdc19adae88f9dc7b4b8caec82b2a2afecc1ecb4b2d2a3a7aeeecfeacd9bf2daeee28dfbc4eaa7c281b49bd884ce92abab95ab9ba59edc83c39c88c2ccb6b891d4999983b29ff5abef92a3d6c0dee3dff391cfc501c6f3bccba3dc8ba5f28ca094a8caa9d5b593d4d9efbdcbcfeb87acc6c299a3a0e696e6fcf290e0eef0d7c4b1f2c486d5aaf9dda383b0baabd59cc9a6ffc4f5f1e4eaf3cbadfdd781819d8887aa90a2b1a7c5a3e8e7aabdf2d4e59a8de6aeb9d394bb84fbade98eda8ef1eaa3a8839782f297c397c285ddf584a9c4f8c1f3e1e7fcefbda897c1eff9b9f592a2c987c3a8d8949bd8ebb792fdaa8f8fe784a8a380c1b0cdd384e5ecccd4c4b5aef6a3f2ccc6a491cac68ec3d8c99daea7dded929fb1cbb8b3c6bdc6bbb7d285d3f9cf9696cb9a919282b2f8b5c7f2abbb9ea0bc8e9ee8b5a29bade185d7998cfbfea5a184b3e9ff83efbcc2e586b0cc93b099c291f98f86df8cceb7cb9dfbb38ec19bd699bdb8888fc5fee2fbf6d9afa5c3e0c1d797bdc071 0xe6a3d5a5bdebdb9f8a82d482d6948a88c7c1f4cfc6cb97e2999acdb4f588a1dab49ee3a2e3c7efa1c4ef95f6e4afb684e8e1b2de87a5da9fb3dfbbf5b5a9b5cf87c0efbfd5d3d8bbe6d4df9f9ca5ae83b3aaaafaf4a0949a98ddf58bb7a690c3f687fabda89793b6dbbff1bef39ae7a0ce8698edd8ab89fa8bd9928dc794a8b3a299ecc5c3f5b1acc6909be8df9ae1959daff5f595e6dbd089ac8f9bbffa8ba8a1b4e0829ca5d7e2f380d3d3adca8de7e8aef3a0dabce2c19cbf94b3f0b6e7b59ce886dcdaa3cb848d8facc3bec9d0bac0df9e88e1a7f183cdf4a2f2fca3bba1eafdc5cfb8b7ebb5e8b7b2c8e0d6e5ddcbad81eedfccdcebc6eeb285c7d7c8caf1d9a9d09ea4e1a0baa2979df0cfa5a2e8f68eb4d8b9fac399d3dbbde6b0a2948fd990af01a9e3cdcbf4f8e2eeac84ec8ade95cec7e6dad1aacf8ffcef96eec3a4ce85e5d7dbb6c789a0ecdcc490e5ccecf1fc88c6dbabd8b8d7e1faea8ecbdcc9d892f1c1f5f2809f83dc89f6cdce8ec0cb81c5be9af68dddc2c8b6dfd6efadbe91a58ea28bbba6b4adc8d89dfdd9a7badbc6e6f8b7a7a1bba4aebeb489ca91cfc5c5faa48f81fb8f9bfd9ae5bd9cbec1c397cbe8eadfeafff6ab939eadad82859fb09681f996b184d1dde1eae1fcc3d08caad1bdeff6d1b7a5f0858a9293deeaa2adb0c7dcd9c5e7abc1d18eacf7b9fcbe87cc83d0c585dc899585b0b3cce6fdabd8d8d495cfe3c5cbd7cef1f3db81d9efafeecfb497d583fceba7d3c882a4cbd6a0dda6b7d4a7dde1a8f89da4d5819ba6e2dbf9ebec8fabd79db88eda8ddda0abd094c7b3ecc8a8022c4b04439efbca5d04f25dadad66bf1b44dcf761e442f8dd000000151bd03ca4d43827976695ea4a09dcfe41b210c99a82' --storage '0x00'
```

where timelock.tz is:
```
storage (bytes);
parameter (pair %time (chest_key) (chest));
code {
       UNPAIR;
       DIP {DROP};
       UNPAIR;
       DIIP {PUSH nat 3600};
       OPEN_CHEST;
       IF_LEFT
         { # successful case
           NIL operation;
           PAIR ;
         }
         {
           IF
             { # first type of failure
               PUSH bytes 0x01;
               NIL operation;
               PAIR;
             }
             { # second type of failure
               PUSH bytes 0x00;
               NIL operation;
               PAIR;
             }
         }
     }
```

### Relevant log output

```shell
Error: Please specify a non-empty storage value in the CLI or in the config file.
┌─────────────┬────────────────────────────┐
│ Contract    │ Result                     │
├─────────────┼────────────────────────────┤
│ timelock.tz │ Bad input or storage value │
└─────────────┴────────────────────────────┘
```

### Screenshots and images

Screen Shot 2022-05-06 at 19 31 43

### Code of Conduct

### LINKED ISSUE #705: [Feature]: allow invocation of Michelson entrypoints with annotations

`taq simulate sample.tz 'Left (Left 3)' --storage 5` will invoke the entrypoint annotated with "decrement".

But we want to be able to do this instead: `taq simulate sample.tz '3' --storage 5 --entrypoint decrement`

### Code of Conduct

### LINKED ISSUE #717: [Feature]: makes debugging quicker

`logPluginRequests` is a flag used for debugging.

e.g. `taq generate types --logPluginRequests` will print out a command used for calling the plugin. It also copies the command to the clipboard for us.

In order to debug, users will have to paste that whole thing into an editor and then append `--inspect-brk` after `node`, and then paste that command to the CLI and run it. Then it'll wait for the user to attach a debugger to the process.

I propose 2 things:

1. Have a short flag for `--logPluginRequests`. Maybe `-lpr`?
2. Instead of copying the command printed out by, let's say `taq generate types --logPluginRequests`, append `--inspect-brk` (maybe after `node`) to the command then copy that new slightly modified command to the clipboard for us.

Alternatively, if `logPluginRequests` is not always used like this, we could have a new flag called `--debug` or `-d` that does number 2.

Makes debugging faster which increases productivity.

### Code of Conduct

jchenche:
Fixes #702
Fixes #705
Fixes #717


tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -262,6 +262,39 @@ describe("E2E Testing for taqueria typechecker and simulator tasks of the tezos-
...
+        await exec(`cp e2e/data/byteSlice.tz ${taqueriaProjectPath}/artifacts`)
+
+        // 2. Run taq simulate
+        const {stdout, stderr} = await exec(`taq simulate byteSlice.tz '0x12a47ef2' --storage '0x00'`, {cwd: `./${taqueriaProjectPath}`})
+
+        expect(stdout).toContain('0xa47ef2')
+        expect(stderr).toBe('');
+    })
+
+    test("Verify that taqueria simulator task can understand Left and Right (part of input and storage value) correctly", async () => {
```
alexzbusko, line 276: Is this test just checking the contents of the smart contract? I think this is something we should do for our tests but I worry about what contracts are getting executed. I would like to talk about how we can cover this if needed. Otherwise we shouldn't be too concerned about the contents of compiled contracts since that isn't something we should be worried about since other people elsewhere work on that

tests/e2e/taqueria-plugin-contract-types.spec.ts
```diff
@@ -123,13 +123,13 @@ describe("E2E Testing for taqueria contract types plugin",  () => {
...
             await exec(`cd ${taqueriaProjectPath} && taq generate types`)
             await exec(`cp e2e/data/timelock.ts ${taqueriaProjectPath}/types`)
-            await exec(`cd ${taqueriaProjectPath}/types &&
-                        npm init -y &&
-                        npm install typescript --save-dev &&
-                        npm install @taquito/taquito &&
-                        npm install @taquito/signer &&
-                        npx tsc timelock.ts`)
-            const {stdout, stderr} = await exec(`cd ${taqueriaProjectPath}/types && node timelock.js`)
+            await exec(`cp e2e/data/tsconfig.timelock.json ${taqueriaProjectPath}/types`)
```
hu3man, line 126: Should we leave these as individual `exec()` statements, or revert to a compound shell command?

jchenche: I think this way, it's easier to debug which command failed.

hu3man (approved):
The core functionality works as expected for me and the fix for the contract types test has resolved the issue locally and in the pipeline

Looks all good on my end but we should get @mweichert and @alexzbusko to review the implementation and test implementation before merging

tests/e2e/taqueria-plugin-contract-types.spec.ts
```diff
@@ -123,13 +123,13 @@ describe("E2E Testing for taqueria contract types plugin",  () => {
...
+            await exec(`npm init -y`, {cwd: `./${taqueriaProjectPath}/types`})
+            await exec(`npm install typescript --save-dev`, {cwd: `./${taqueriaProjectPath}/types`})
```
alexzbusko, line 128: @jchenche this looks good to me, the only thing I am wondering is if we can move these dependencies to the tests package.json so that they don't need to be installed during the test and can be used by everything in the test folder. I know the package.json already includes the @taquito/taquito package and typescript. This might cut down the number of commands in this test

alexzbusko (approved):
LGTM, comment is something that can be looked into but also left for the future

tests/e2e/taqueria-plugin-contract-types.spec.ts
jchenche: Done!

hu3man (approved):
🚀

hu3man:
@mweichert, please review asap so we can get this merged 🙏

plugins.ts
```diff
@@ -35,10 +35,13 @@ export const inject = (deps: PluginDeps) => {
         attemptP(async () => {
             if (parsedArgs.logPluginRequests)  {
                 const encoder = new TextEncoder()
-                const output = pluginRequestToString (plugin) (cmd)
+                let output = pluginRequestToString (plugin) (cmd)
                 await stdout.write(encoder.encode(`*** START Call to ${plugin.name} ***\n`))
                 await stdout.write(encoder.encode(`${output}\n`))
                 await stdout.write(encoder.encode(`*** END of call to ${plugin.name} ***\n`))
+                if (parsedArgs.debug) {
```
mweichert, line 42: Jimmy, we discussed meeting to review how to do this in a way that avoids a SoC violation.

Do you mind doing the following please:
1) Create an issue to fix the SoC violation here
2) Add a TODO comment that links to the issue.

Thanks!

taqueria-plugin-tezos-client/client.ts
```diff
@@ -160,19 +160,21 @@ const preprocessString = (value: string) : string => {
 
 const getSimulateCommand = (opts: Opts, sandbox: Sandbox, sourceFile: string, sourcePath: string) => {
     let storage
```
mweichert, line 162: Jimmy, as you know, I'm not a big fan of using `let` as it makes the code harder to read/reason-about.

Here's how I would suggest refactoring this (untested code below):
```ts
const storageGiven  = opts.storage ?? getStorageFromConfig(opts, sourceFile)
const inputGiven = opts.input ? opts.input.toString() : undefined

if (!storageGiven) throw new Error('Error: Please specify a non-empty storage value in the CLI or in the config file.')
if (!inputGiven) throw new Error('Error: Please specify an input value.')

const storage = preprocessString(storageGiven!)
const input = preprocessString(inputGiven)
const entrypoint = opts.entrypoint ? `--entrypoint ${opts.entrypoint}` : ''
const cmd = `docker exec ${sandbox.name} tezos-client run script ${sourcePath} on storage \'${storage}\' and input \'${input}\' ${entrypoint}`

return cmd
```

cli.ts
```diff
@@ -541,9 +541,13 @@ const executingBuiltInTask = (inputArgs: SanitizedInitArgs | RawInitArgs) =>
             
 export const run = (env: EnvVars, inputArgs: DenoArgs, i18n: i18n) => {
     try {
-        // A hack to get around yargs because it strips leading and trailing double quotes of strings passed by the command
-        // Refer to https://github.com/yargs/yargs-parser/issues/201
-        inputArgs = inputArgs.map(arg => arg.match(/^"(.|\n)*"$/) ? "___" + arg + "___" : arg)
+        inputArgs = inputArgs.map(arg => {
```
mweichert, line 544: Would you mind moving these hacks to a separate function? Perhaps call it `preprocessArgs` or something like that.

taqueria-sdk/index.ts
```diff
@@ -146,9 +146,29 @@ const sanitizeArgs = (parsedArgs: ParsedArgs) : Promise<SanitizedArgs> =>
 
 const parseArgs = (unparsedArgs: Args): LikeAPromise<ParsedArgs, Failure<undefined>> => {
     if (unparsedArgs && Array.isArray(unparsedArgs) && unparsedArgs.length >= 2) {
+
```
mweichert, line 149: Mind moving the hacks to a separate function, such as `preprocessArgs` please? Isolates some of the noice that the hacks add to this function.

tests/e2e/data/tsconfig.timelock.json
```diff
@@ -0,0 +1,101 @@
...
+    /* Projects */
+    // "incremental": true,                              /* Enable incremental compilation */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "ES6",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
```
mweichert, line 14: Can we change this to ES2021 please?

mweichert (changes requested):
@jchenche thanks for taking this on. Overall, looks good - I just added some rather small requests. Let me know if you have any questions.

plugins.ts
jchenche: Got it! The issue is here: https://github.com/ecadlabs/taqueria/issues/732

Will push a commit to contain this TODO comment.

taqueria-plugin-tezos-client/client.ts
jchenche: Will refactor this to eliminate `let` 👍

taqueria-plugin-tezos-client/index.ts
```diff
@@ -45,6 +45,11 @@ Plugin.create(i18n => ({
                     description: "The name of the sandbox to use",
                     required: false
                 }),
+                Option.create({
```
hu3man, line 48: We should update the documentation to add this functionality. What would the command look like using this?
`taq simulate file.tz `10` --entrypoint '%increment'`

jchenche: `taq simulate sample.tz '2' --storage '3' --entrypoint increment`

mweichert (approved):

PR #734, opened 2022-05-19
Added dependency injection to vscode extension

### LINKED ISSUE #733: Create a mocking mechanism to VS Code Extension
As a Developer, I want to add a mocking mechanism to VS Code Extension so SDET can use it to mock responses running a plugin command.
Use case:  Running taq init command inside VS Code opens the dialogue box to select the folder that the automation test cannot bypass.  Need a solution to return a different response (skip opening dialogue box) to be able to test it. As a result, the test will focus on verifying mocked responses.

mweichert:
Fixes #733


sinapsist (approved):
LGTM

jchenche (approved):

PR #779, opened 2022-05-27
Jazz up CLI output

hu3man:


mweichert:
Looking good thus far @hu3man !

PR #843, opened 2022-06-06
Various fixes to jest plugin

mweichert:
- Fixes testPattern not matching against existing files
- Fixes buffering of Jest output
- Ensures that stdout and stderr are used appropriately
- Fixes issue with colourized output lost from jest


hu3man (approved):
LGTM

PR #850, opened 2022-06-07
VSCode: Removed "debug mode" - added plugin list override method

mweichert:
Create taqueria-plugins.json with an array of paths to override path


another-test/quickstart.md
```diff
@@ -0,0 +1,215 @@
+## Overview
```
alexzbusko, line 1: Should this be here?

hu3man (approved):

PR #852, opened 2022-06-08
Fixes #572, adds jest test stubs

### LINKED ISSUE #572: Test Stub Generation
As a test author, I want Taqueria to generate test stubs for all entry points in my project's contracts so that I save time and increase coverage when writing tests

As a type generation plugin user I want to have test stubs for all contract entry points be automatically generated as part of the type generation plugin so that can I have test coverage across all entry points.

***USE CASE***
The user using the type generation plugin generates type information for their smart contract and at the same time is provided with test stubs for the entry points within the contract.

mweichert:
Working test stub generation. Missing tests.

## Documentation notes:

The jest plugin now exposes a template called _contract-test_. Like any other template, you would use the _create_ task to instantiate it:
```
taq create contract-test [michelsonArtifact]
```

When instantiated, it will create a jest test suite for you with some test cases used to test the functionality of your contract. The test suite depends on taqutio, as well as Rick's Contract Types plugin to generate types for the contract.

### Options
The _contract-test_ template supports the following options:
*--name* - used to name the file of the generated test suite. Defaults to the basename of the artifact name.
*--partition* - the name of the testing partition to store the test suite. Defaults to jest's _testsRootDir_ configuration setting.

### Example

Run: `taq create contract-test increment.tz`

Expected results:
- _tests/types_ will include generated types for the contract
- _tests/increment.ts_ is the newly generated test suite

### Plugin Runtime Dependencies

At this time, plugins do not have a way of declaring their runtime dependencies or providing a means of installing them. In this particular case, the jest plugin should declare the following runtime dependencies for an optimal end-developer experience:
- @types/jest
- @taquito/taquito
- @taquito/utils
- ts-jest

Without the end-developer installing these, the user will be unable to run tests.


hu3man:
If there are any nuances to the feature other than the obvious (run task -> output stubs), please let me know so I can document it

PR #863, opened 2022-06-09
fix: fix the failing assertion when the test project is initialized

GImbrailo:
In the `main.yml` workflow there is an assertion made after the binaries are compiled:

```shell
[[ $(./"${{ matrix.taqueria_bin }}" init -p ./test_project) == "Project taq'ified!" ]] 
```
The assertion no longer works on the windows runner. This PR aims to fix that by making a new assertion that checks if the `config.json` file exists in the `test_project/.taq/` directory


danielelisi (approved):

jchenche (approved):

hu3man (approved):
Awesome, thanks @GImbrailo !

Issue #565, opened 2022-04-07
As a TVSCE user, I want the command palette options to include ONLY and ALL valid tasks/parameteers so that I'm not confused about what I can/can't run in the current project


hu3man:
### User story

- As a TVSCE user, I want the command palette options to include ONLY and ALL valid tasks/parameteers so that I'm not confused about what I can/can't run in the current project
- As a taqueria developer, I want the command palette options to be dynamically populated so that I don't have to manually update the VSCE codebase every time there is a change to a plugin or the CLI that modifies tasks

### Is this feature solving a problem?

Currently, the available taq commands are hard coded lists. This results in an increased maintenance burden for the team as well as confusion for the developer as they are presented commands for plugins that are not installed on the project

### Please share any additional details you have?

**Acceptance Criteria**

- [ ] Opening the command palette and typing `taq` should return the full list of commands available to run based on the plugins installed on the current project 
- [ ]  If the current project is not initialized, the user should be presented the scaffold, and init options
- [ ]  For `taq originate`, a secondary menu should open which presents the user a list of environments configured on the project, then uses the `--env [envName` flag to target the Origination properly
- [ ] 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #864, opened 2022-06-10
show/hide & enable/disable tasks

AlirezaHaghshenas:
closes: #565

### DIFF
--- a/taqueria-vscode-extension/src/extension.ts
+++ b/taqueria-vscode-extension/src/extension.ts
@@ -13,6 +13,7 @@ export async function activate(context: api.ExtensionContext, input?: InjectedDe
 		exposeInstallTask,
 		exposeTaqTaskAsCommand,
 		exposeSandboxTaskAsCommand,
+		updateCommandStates,
 	} = inject(deps);
 
 	const i18n: i18n = await loadI18n();
@@ -33,7 +34,7 @@ export async function activate(context: api.ExtensionContext, input?: InjectedDe
 	// Third-party plugins aren't exposed via the VS Code interface
 	if (folders.length === 1) {
 		await makeDir(folders[0].uri.path, i18n)
-			.then(projectDir => {
+			.then(async projectDir => {
 				const exposeTaqTask = exposeTaqTaskAsCommand(context, output, i18n, projectDir);
 				const exposeSandboxTask = exposeSandboxTaskAsCommand(context, output, i18n, projectDir);
 
@@ -53,6 +54,10 @@ export async function activate(context: api.ExtensionContext, input?: InjectedDe
 
 				// Originate task
 				exposeTaqTask(COMMAND_PREFIX + 'deploy', 'deploy', 'output', 'Deployment successful.');
+
+				setTimeout(async () => {
+					await updateCommandStates(context, output, i18n, projectDir);
+				}, 3000);
 			});
 	}
 
--- a/taqueria-vscode-extension/src/extension.ts
+++ b/taqueria-vscode-extension/src/extension.ts
@@ -1,11 +1,38 @@
 import loadI18n, { i18n } from '@taqueria/protocol/i18n';
 import path from 'path';
 import * as api from 'vscode';
-import { clearFileSystemWatchers, inject, InjectedDependencies, sanitizeDeps } from './lib/helpers';
+import { inject, InjectedDependencies, sanitizeDeps } from './lib/helpers';
 import { COMMAND_PREFIX } from './lib/helpers';
 import { makeDir } from './lib/pure';
 
-const watchers = new Map<string, api.FileSystemWatcher>();
+const { clearConfigWatchers, getConfigWatchers, addConfigWatcherIfNotExists } = (() => {
+	const inMemoryState = {
+		configWatchers: new Map<string, api.FileSystemWatcher>(),
+	};
+
+	const clearConfigWatchers = () => {
+		for (const watcher of inMemoryState.configWatchers.values()) {
+			watcher.dispose();
+		}
+		inMemoryState.configWatchers.clear();
+	};
+
+	const getConfigWatchers = () => inMemoryState.configWatchers.values();
+
+	const addConfigWatcherIfNotExists = (folder: string, factory: () => api.FileSystemWatcher) => {
+		if (inMemoryState.configWatchers.has(folder)) {
+			return;
+		}
+		const watcher = factory();
+		inMemoryState.configWatchers.set(folder, watcher);
+	};
+
+	return {
+		clearConfigWatchers,
+		getConfigWatchers,
+		addConfigWatcherIfNotExists,
+	};
+})();
 
 export async function activate(context: api.ExtensionContext, input?: InjectedDependencies) {
 	const deps = sanitizeDeps(input);
@@ -36,7 +63,7 @@ export async function activate(context: api.ExtensionContext, input?: InjectedDe
 	// Third-party plugins aren't exposed via the VS Code interface
 	if (folders.length === 1) {
 		await makeDir(folders[0].uri.path, i18n)
-			.then(async projectDir => {
+			.then(projectDir => {
 				const exposeTaqTask = exposeTaqTaskAsCommand(context, output, i18n, projectDir);
 				const exposeSandboxTask = exposeSandboxTaskAsCommand(context, output, i18n, projectDir);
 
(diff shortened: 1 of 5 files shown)


hu3man (commented):
Hey @AlirezaHaghshenas, this looks like a great solution to me and seems to be working quite well in my manual testing for the most part!

The only issue I see is that we should not display `install plugin` when not in a Taqueria project. It seems to just fail silently if it's attempted

For the best UX, we should  not show/allow users to install plugins on non-taq'ifed dir's. If this is not possible of difficult, providing a helpful error message if the user does attempt to install plugins on a non-initialized dir would be helpful

hu3man (approved):
🚀

taqueria-vscode-extension/src/extension.ts
```diff
@@ -33,7 +35,7 @@ export async function activate(context: api.ExtensionContext, input?: InjectedDe
 	// Third-party plugins aren't exposed via the VS Code interface
 	if (folders.length === 1) {
 		await makeDir(folders[0].uri.path, i18n)
-			.then(projectDir => {
+			.then(async projectDir => {
```
jchenche, line 38: Why did we add `async` when there's no `await` in the function?

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -328,6 +332,54 @@ export const inject = (deps: InjectedDependencies) => {
...
+	const updateCommandStates = async (
+		context: api.ExtensionContext,
+		output: api.OutputChannel,
+		i18n: i18n,
+		projectDir: Util.PathToDir,
+	) => {
+		try {
+			const config = await Util.TaqifiedDir.create(projectDir, i18n);
+			vscode.commands.executeCommand('setContext', '@taqueria-state/is-taqified', !!config.config);
+			for (let plugin of getWellKnownPlugins()) {
```
jchenche, line 344: Is it possible to use `const` instead of `let`?

AlirezaHaghshenas:
> Hey @AlirezaHaghshenas, this looks like a great solution to me and seems to be working quite well in my manual testing for the most part!
> 
> The only issue I see is that we should not display `install plugin` when not in a Taqueria project. It seems to just fail silently if it's attempted
> 
> For the best UX, we should not show/allow users to install plugins on non-taq'ifed dir's. If this is not possible of difficult, providing a helpful error message if the user does attempt to install plugins on a non-initialized dir would be helpful

Yes, that's updated in the next commit. Thanks

taqueria-vscode-extension/src/extension.ts
AlirezaHaghshenas: Yes, this is a remnant of an older change. Now that nothing is async, I can revert this change.

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -328,6 +330,46 @@ export const inject = (deps: InjectedDependencies) => {
...
+	};
+
+	const createWatcherIfNotExists = (
+		context: api.ExtensionContext,
+		output: api.OutputChannel,
+		i18n: i18n,
+		projectDir: Util.PathToDir,
+		createConfigWatcherIfNotExists: (folder: string, factory: () => api.FileSystemWatcher) => void,
+	) => {
+		createConfigWatcherIfNotExists(projectDir, () => {
```
jchenche, line 359: This is a function passed as an argument to `createWatcherIfNotExists`

1. Why is `createConfigWatcherIfNotExists` an argument? Can't it be a normal function?

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -328,6 +330,46 @@ export const inject = (deps: InjectedDependencies) => {
...
 
+	const updateCommandStates = async (
+		context: api.ExtensionContext,
+		output: api.OutputChannel,
+		i18n: i18n,
+		projectDir: Util.PathToDir,
+	) => {
+		try {
+			const config = await Util.TaqifiedDir.create(projectDir, i18n);
+			vscode.commands.executeCommand('setContext', '@taqueria-state/is-taqified', !!config.config);
```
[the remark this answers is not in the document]
mweichert: @jchenche it's a double negative - equivalent to `config.config ? true : false`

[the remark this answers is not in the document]
jchenche: Oh so it's not true that `config.config` the same as `!!config.config`?

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -328,6 +330,46 @@ export const inject = (deps: InjectedDependencies) => {
...
+			const plugins = getWellKnownPlugins();
+			for (const plugin of plugins) {
+				const found = config.config.plugins?.find(item => item.name === plugin) !== undefined;
+				vscode.commands.executeCommand('setContext', plugin, found);
+			}
+		} catch (e: any) {
+			// After logging PR is merged, log this error
```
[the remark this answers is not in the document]
mweichert: @AlirezaHaghshenas this shouldn't hold back an approval, but I thought I'd make a suggestion here.

One thing about loops is that the reviewer/readable has to read the code to understand the logic.

To make that easier, its get to have a single-purpose pure functions to help convey the logic.

E.g.:

```
const isPluginInstalled = (config: Config.t) => (pluginName: string) => config.plugins?.find(item => item.name === pluginName) || false

const installedPlugins = config.plugins?.filter(isPluginInstalled) || false

const activatePlugins = (installedPlugins) => 
  installedPlugins.map(plugin => vscode.commands.executeCommand('setContext', plugin.name, plugin)  
```

I'd argue that's more readable as I'm using domain language - isPluginInstalled, installedPlugins, activatePlugins. Anyhow, just a suggestion going forward. It has the advantage that some of those can exist in pure.ts as helper methods.

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -328,6 +330,46 @@ export const inject = (deps: InjectedDependencies) => {
...
+		projectDir: Util.PathToDir,
+		createConfigWatcherIfNotExists: (folder: string, factory: () => api.FileSystemWatcher) => void,
+	) => {
+		createConfigWatcherIfNotExists(projectDir, () => {
+			const watcher = vscode.workspace.createFileSystemWatcher(join(projectDir, '.taq/config.json'));
+			// TODO: We should detect the event that VsCode's current Folder is changed and the watcher should be disposed
+
+			updateCommandStates(context, output, i18n, projectDir);
+
+			// TODO: Is passing these arguments to the callback of a long lived watcher prevent GC? Are these short lived objects?
```
mweichert, line 365: Indeed, as they're in closure scope, they will be long lived. 

As an alternative, I'd say adjust updateCommandStates to accept an object of key/pair values instead:
```
updateCommandStates({
  context,
  output,
  i18n,
  projectDir
})
```

Saying that, many of those will just be references so its probably not a significant issue.

mweichert (approved):
LGTM. I did make some suggestions, but nothing that should hold back the PR.

taqueria-vscode-extension/src/lib/helpers.ts
AlirezaHaghshenas: Because that functionality is implemented in `extension.ts`. That file depends on `helpers.ts`. If I call that method directly, I will also depend `helpers.ts` on `extension.ts`. Regardless of whether or not this is supported by our project system, this is a cyclic dependency that is better be avoided.

jchenche (approved):

taqueria-vscode-extension/src/lib/helpers.ts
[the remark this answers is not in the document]
AlirezaHaghshenas: `!!something` will always be `true` or `false`, while `something` might be anything that's `truthy` or `falsy`.

taqueria-vscode-extension/src/lib/helpers.ts
AlirezaHaghshenas: In general passing functions as arguments is not a bad thing. Lots of functionality like `map` and `filter` take functions as arguments. But whether or not this specific implementation could be done better can be a good water-cooler conversation for knowledge sharing.

jchenche: Yes but this argument seems very specific, just by looking at the name: `createConfigWatcherIfNotExists`. I am under the impression that only one function value will ever be passed to `createWatcherIfNotExists`. If that's the case, it sounds like a code smell.

PR #870, opened 2022-06-14
Fix Ligo docker artifacts permission problem

AlirezaHaghshenas:


hu3man (approved):

taqueria-plugin-ligo/compile.ts
```diff
@@ -28,7 +28,7 @@ const getCompileCommand = (opts: Opts) =>
 
 		const inputFile = getInputFilename(opts)(sourceFile);
 		const baseCommand =
-			`DOCKER_DEFAULT_PLATFORM=linux/amd64 docker run --rm -v \"${projectDir}\":/project -w /project ligolang/ligo:next compile contract ${inputFile}`;
+			`DOCKER_DEFAULT_PLATFORM=linux/amd64 docker run --rm -v \"${projectDir}\":/project:z -w /project ligolang/ligo:next compile contract ${inputFile}`;
```
jchenche, line 31: How does adding `:z` fix the problem?

AlirezaHaghshenas: It does not. Sorry, I have marked the PR as draft.
The `:z` is useful anyway, for SELinux scenarios, I just wanted to push, to see what happens to all tests as a result of this change.

AlirezaHaghshenas:
The reason for this problem is: the docker is run as root, so the artifacts it creates are owned by root.
The solution is to use `-u` to map the host user to container user, but it fails with Ligo image. I'll be trying to find another solution.

hu3man:
@AlirezaHaghshenas, can we convert this to draft till it's ready for review?

jchenche (approved):

PR #875, opened 2022-06-20
415 linux permissions tests

alexzbusko:
Tests to ensire the correct permissions are being set for different files produced by taqueria


mweichert (approved):
Good stuff here, Alex. 

One thing to note - if ever did want to support Windows, we'd need to adjust the exec call to use the equivalent functions from the `fs` module: https://nodejs.org/dist/latest-v16.x/docs/api/fs.html

For instance, instead of using exec to call `cp`, `stat`, etc we'd use `fs.copy`, `fs.stat`.

In the future, it might be a better practice for us to use the `fs` module instead of exec unless really needed to keep the code portable.

For instance, instead of

tests/e2e/taqueria-cli-permissions.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+		userGroup = (await exec(`id -g -n ${username}`)).stdout.trim();
+	});
+
+	test('testing that ligo artifacts will have the correct permissions', async () => {
+		await exec(`taq compile --plugin ligo`, { cwd: `./${taqueriaProjectPath}` });
+		const fileUser = await exec(`stat -c %U ${taqueriaProjectPath}/artifacts/increment.tz`);
+		const fileGroup = await exec(`stat -c %G ${taqueriaProjectPath}/artifacts/increment.tz`);
+
+		expect(fileUser.stdout.trim()).toBe(username);
+		expect(fileGroup.stdout.trim()).toBe(userGroup);
```
AlirezaHaghshenas, line 29: If the user running these tests is `root`, (Is it? in our CI pipelines?) then these tests will pass anyway, even if the functionality being tested sets permissions to `root` not to the current user.
TBH, I don't have a suggestion to fix this.

PR #882, opened 2022-06-21
Fixes #876, limits the number of task outputs which are persisted

### LINKED ISSUE #876: Limit the number of task outputs persisted
We made an adjustment to save task outputs to the persistent state / state registry along with the output of operations. This was to solve an immediate need - have the ability to reference the contract address of the last origination.

Although I'd treat this as an experimental feature that could be dropped at some point, until the state API has been fully realized this is a good feature to offer in the midterm.

However, _all_ task outputs are persisted, meaning that the persistent state can grow large and quickly. What would be best is to add a configurable limit per task in terms of how many outputs are persisted.

I think the last 5 outputs is a reasonable default.

In the config file, the end-developer can adjust the default:

```json
"state": {
  "tasks": {
     "limit": 5,
     "originate": 10
  }
}
```

The above configuration would ensure that the last 5 outputs of any task are persisted, with the `originate` increased to the last 10 outputs.

mweichert:
Later we'll want to add a configurable limit per task type, but for now I've just hardcoded a limit that will retain the latest 5 outputs.


test-project/quickstart.md
```diff
@@ -0,0 +1,209 @@
+## Overview
```
alexzbusko, line 1: Does this need to be here?

mweichert: Nope. Removed. :)

hu3man (approved):
This seems reasonable for originations. We May need to revisit as more stateful operations are added

PR #902, opened 2022-06-25
Fixes #901, use provisioner.ts file for provisions with DSL that provides code-completion

### LINKED ISSUE #901: As a taq end-developer, I would like an API with code completion available in the provisions.ts file
Scope of this task:
- [x] Adjust CLI to create a .taq/provisioner.ts file instead of a .taq/provisioner.json file
- [x] Dynamically generate TypeScript declaration for .taq/provisioners.ts file

The API should resemble the following:
```
export const p10 = 
  provision("compile-hello.mligo")
  .task(state => tasks.ligo.compile({
     contract: "hello.mligo"
   })
  .when(state => state["hello.mligo"].hasChanged())

export const p20 = 
  provision("originate-hello.mligo")
  .task(state => tasks.taquito.originate({
     contract: state["hello.mligo"].artifactName
   })
  .after([p10])
```

The `tasks` property should infer the list of plugins and tasks available from the ephemeral state.

The state object should infer which contracts have been registered by referencing the config.json file.

mweichert:
This PR does the following:
- Replace provisioners.json with provisioner.ts
- Dynamically generate type declaration file using ephemeral state, persistent state, and the config.json file.


mweichert:
Still need to do the following:
- Add the following to the provisioners.ts file when first generated: `/// <reference types="./provisioner.d.ts" />`
- Ensure that persistent state files are created before trying to generate type declarations

tests/e2e/taqueria-bootstrap.spec.ts
```diff
@@ -0,0 +1,182 @@
+import { spawnSync } from 'child_process';
+import { readdir, readFile, rm, writeFile } from 'fs/promises';
+import { join } from 'path';
+import { generateTestProject } from './utils/utils';
+
+const taqueriaProjectPath = 'e2e/taq-bootstrap';
+
+const expectedConfig = `
```
alexzbusko, line 8: Since these are just strings you can always put them into a data file and then import them for this test just to cut down on what is in this file

tests/e2e/taqueria-bootstrap.spec.ts
```diff
@@ -0,0 +1,182 @@
+import { spawnSync } from 'child_process';
```
alexzbusko, line 1: Is there a particular reason you're using spawnSync vs the promisified exec method used in other spec files? One reason I ask is because I have noticed that mixing synchronous and asynchronous stuff has lead to some stability issues when running in the pipeline. If possible I want to keep things async in our tests

taqueria-protocol/ParsedPluginInfo.ts
```diff
@@ -1,29 +1,10 @@
...
-	operations: z.preprocess(
-		val => val ?? [],
-		z.array(
-			ParsedOperation.schemas.schema,
-			{ description: 'ParsedOperations' },
-		)
-			.optional(),
-	),
-}).describe('ParsedPluginInfo');
+const internalSchema = PluginInfo.internalSchema.extend({}).describe('ParsedPluginInfo');
```
jchenche, line 5: Why are we deleting or trimming all these `zod` types in this PR?

jchenche (approved):
Only have some questions about some of the changes in this PR. Other than that 👍

cli.ts
```diff
@@ -462,27 +464,27 @@ const addOperations = (
...
 			() => {},
 			(argv: Arguments) =>
 				pipe(
 					SanitizedArgs.of(argv),
-					map(inputArgs => joinPaths(inputArgs.projectDir, '.taq', 'provisions.json')),
-					chain(SanitizedAbsPath.make),
-					chain(loadProvisions),
-					map(plan),
-					forkCatch(displayError(cliConfig))(displayError(cliConfig))(log),
+					// map(inputArgs => joinPaths(inputArgs.projectDir, '.taq', 'provisions.json')),
```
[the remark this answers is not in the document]
mweichert: Good catch. I've removed the commented out lines.

cli.ts
```diff
@@ -462,27 +464,27 @@ const addOperations = (
 		(argv: Arguments) =>
 			pipe(
 				SanitizedArgs.ofProvisionTaskArgs(argv),
-				chain(inputArgs => addNewProvision(inputArgs, config, state)),
-				map(() => 'Added provision to .taq/provisions.json'),
-				forkCatch(displayError(cliConfig))(displayError(cliConfig))(log),
+				// chain(inputArgs => addNewProvision(inputArgs, config, state)),
```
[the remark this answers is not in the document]
mweichert: Good catch. I've removed the commented out lines.

taqueria-protocol/ParsedPluginInfo.ts
mweichert: We're removing operations. They're hard to differentiate from tasks, and I've found the real underlying difference is context. So instead of having a separate construct/type called "operation", I'm reusing tasks but in a different context.

tests/e2e/taqueria-bootstrap.spec.ts
mweichert: > I have noticed that mixing synchronous and asynchronous stuff has lead to some stability issues when running in the pipeline.

Didn't realize that. 

> If possible I want to keep things async in our tests

Sounds good! All for it.

> Is there a particular reason you're using spawnSync vs the promisified exec method used in other spec files?

I spent some time testing exec, spawn, and execa (an NPM package) in various environments. I found that spawn is more reliable than exec as you can specify `{shell: true}` in the opts.

execa is actually the best of them all, but its an ESM module which our testing framework doesn't seem to support at this time. I did actually get it working in a separate PR. We can chat about this more on Monday if you'd like.

tests/e2e/taqueria-bootstrap.spec.ts
mweichert: Sounds like this is a preference, which I'm perfectly fine adopting. Moved to e2e/data/bootstrap-contents.ts.

PR #908, opened 2022-06-28
Fixes #906, Added ability to register contracts

### LINKED ISSUE #906: As a taq end-dev, I would like to register contracts so that I may later reference them by name

Currently, Taqueria expects that all contract source files live in _contracts/_.

This works well for the most part - you can run `taq compile`, which will iterate over every source file and compile each one, or `taq compile [sourceFile]` to compile an individual file.

However, this breaks down when a contract is composed of multiple source files - as in that case, not every source file in _contracts_ is also a contract. This is a requirement for https://github.com/ecadlabs/taqueria/issues/234

Each compiler has its own require/include mechanism, its difficult to handle this in a scalable way without putting a lot more responsibility on the plugin author.

As such, we've decided to solve this by introducing the concept of a _registered contract_. Instead of `taq compile` iterating over each source file, it iterates over the list of registered contracts.

We'll need some CR[U]D tasks for managing these registered contracts:
`taq add-contract`
`taq remove-contract`
`taq list-contracts`

The schema for a contract has just two fields:
- sourceFile, a relative path to _contractsDir_.
- hash, a hash of file contents

As a hash is a something that gets computed and then stored we don't need a task to update registered contracts.

It allows us to handle multi-file contracts.

Related issues: 
https://github.com/ecadlabs/taqueria/issues/234

### Code of Conduct

mweichert:
Fixes #906

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -22,6 +22,7 @@ import type { Arguments } from 'https://deno.land/x/yargs@v17.4.0-deno/deno-type
 import yargs from 'https://deno.land/x/yargs@v17.4.0-deno/deno.ts';
 import { __, match } from 'https://esm.sh/ts-pattern@3.3.5';
 import * as Analytics from './analytics.ts';
+import { addContract } from './contracts.ts';
 import * as NPM from './npm.ts';
 import { addTask } from './persistent-state.ts';
 import inject from './plugins.ts';
@@ -309,6 +310,28 @@ const postInitCLI = (cliConfig: CLIConfig, env: EnvVars, args: DenoArgs, parsedA
 				default: false,
 				boolean: true,
 			})
+			.command(
+				'add-contract <sourceFile>',
+				i18n.__('addContractDesc'),
+				(yargs: Arguments) => {
+					yargs.positional('sourceFile', {
+						describe: i18n.__('addSourceFileDesc'),
+						type: 'string',
+						required: true,
+					});
+
+					yargs.option('contractName', {
+						alias: ['name', 'n'],
+						type: 'string',
+					});
+				},
+				(inputArgs: Record<string, unknown>) =>
+					pipe(
+						SanitizedArgs.ofManageContractArgs(inputArgs),
+						chain(args => addContract(args, i18n)),
+						forkCatch(displayError(cliConfig))(displayError(cliConfig))(log),
+					),
+			)
 			.demandCommand(),
 		extendCLI(env, parsedArgs, i18n),
 	);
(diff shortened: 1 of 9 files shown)


test-project/quickstart.md
```diff
@@ -0,0 +1,209 @@
+## Overview
```
hu3man, line 1: Test project snuck in

cli.ts
```diff
@@ -309,6 +310,61 @@ const postInitCLI = (cliConfig: CLIConfig, env: EnvVars, args: DenoArgs, parsedA
 				default: false,
 				boolean: true,
 			})
+			.command(
+				'add-contract <sourceFile>',
```
hu3man, line 314: Hyphenation here breaks from our current standard with space-separated words (`taq list accounts`, `taq start sandbox`, etc)

hu3man (commented):
Looking great @mweichert! 

Left a comment regarding the usage of hyphens in the commands and aliases

Also, it seems that the new commands disappear from the printed help menu and the error displayed is unhelpful when `taq list-contracts` is run when the registry is empty: 
```
taqueria/registerTest on  906-registered-contracts 
❯ taq list-contracts                      
taq [command]

Commands:
  taq init [projectDir]                     Initialize a new project
  taq opt-in                                Opt-in to sharing anonymous usage an
                                            alytics
  taq opt-out                               Opt-out of sharing anonymous usage a
                                            nalytics
  taq scaffold [scaffoldUrl] [scaffoldProj  Generate a new project using pre-mad
  ectDir]                                   e scaffold

Options:
      --version     Show version number                                [boolean]
      --build       Display build information about the current version[boolean]
  -p, --projectDir  Path to your project directory               [default: "./"]
  -e, --env         Specify an environment configuration

Taqueria is currently in BETA. You've been warned. :)

Cannot convert undefined or null to object
```

cli.ts
jchenche: I remember having discussions about turning all of these space-separated words into hyphenated ones because `taq start sand` (should fail but don't) works the same as `taq start sandbox` currently.

cli.ts
```diff
@@ -309,6 +310,61 @@ const postInitCLI = (cliConfig: CLIConfig, env: EnvVars, args: DenoArgs, parsedA
...
+			.command(
+				'list-contracts',
+				i18n.__('listContractsDesc'),
+				() => {},
+				(inputArgs: Record<string, unknown>) =>
+					pipe(
+						SanitizedArgs.of(inputArgs),
+						chain(args => listContracts(args, i18n)),
+						map(renderTable),
+						forkCatch(displayError(cliConfig))(displayError(cliConfig))(identity),
```
jchenche, line 364: Do we have to do `forkCatch` here (instead of doing it at the end of the `run` function)? Is this a `yargs` limitation again?

contracts.ts
```diff
@@ -0,0 +1,104 @@
...
+import { getConfig } from './taqueria-config.ts';
+import { joinPaths, readTextFile, writeJsonFile } from './taqueria-utils/taqueria-utils.ts';
+
+type contractRow = {
+	'Name': string;
+	'Last Known Hash': string;
+	'Source file': string;
+};
+
+const hasContracts = (config: LoadedConfig.t) => isEmpty(config.contracts);
```
jchenche, line 21: Isn't it supposed to be `isNotEmpty` or something like that? Since we are testing if the contracts folder has content?

contracts.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const isContractRegistered = (contractName: string, config: LoadedConfig.t | Config.t) =>
+	config.contracts && config.contracts[contractName] ? true : false;
```
jchenche, line 24: I believe the `&&` operator already turns `true` or `false` so `config.contracts && config.contracts[contractName]` is the same right?

contracts.ts
```diff
@@ -0,0 +1,104 @@
...
+			return writeJsonFile(joinPaths(parsedArgs.projectDir, '.taq', 'config.json'))(updatedConfig);
+		}),
+		chain(_ => listContracts(parsedArgs, i18n)),
+	);
+
+export const listContracts = (parsedArgs: SanitizedArgs.t, i18n: i18n.t) =>
+	pipe(
+		getConfig(parsedArgs.projectDir, i18n),
+		map(config =>
+			hasContracts(config)
```
jchenche, line 93: Do you mean `hasNoContracts(config)`?

cli.ts
hu3man: That is actually a good failure, we want the cli to be as resilient to user error as possible

hu3man:
@mweichert, are you able to provide some high-level notes/instructions for using contract registration that I can start writing documentation off of?

alexzbusko (approved):
LGTM

cli.ts
mweichert: Buit-in functions use forkCatch as they don't require plugins to be loaded.

contracts.ts
mweichert: Yep, fixed.

PR #910, opened 2022-06-28
Adding more logs for the command state watcher

AlirezaHaghshenas:


taqueria-vscode-extension/src/extension.ts
```diff
@@ -19,11 +19,14 @@ const { clearConfigWatchers, getConfigWatchers, addConfigWatcherIfNotExists } =
 
 	const getConfigWatchers = () => inMemoryState.configWatchers.values();
 
-	const addConfigWatcherIfNotExists = (folder: string, factory: () => api.FileSystemWatcher) => {
+	const addConfigWatcherIfNotExists = (folder: string, factory: () => api.FileSystemWatcher | null) => {
 		if (inMemoryState.configWatchers.has(folder)) {
 			return;
 		}
 		const watcher = factory();
+		if (watcher == null) {
```
mweichert, line 27: Is there a way we could avoid returning null here? Intuitively, I would think that a factory would either return a new object or throw.

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -276,6 +277,45 @@ export const inject = (deps: InjectedDependencies) => {
 		vscode.window.showInformationMessage(msg)
 			.then(_ => Promise.resolve()) as Promise<void>;
 
+	const logAllNestedErrors = (err: TaqVsxError | TaqError | Error | any, output: Output) => {
```
mweichert, line 280: Is it possible to avoid the `any` here please? 

In general, I'd like to avoid using the `any` type in the codebase.

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -391,19 +436,30 @@ export const inject = (deps: InjectedDependencies) => {
 		output: Output,
 		i18n: i18n,
 		projectDir: Util.PathToDir,
-		addConfigWatcherIfNotExists: (folder: string, factory: () => api.FileSystemWatcher) => void,
+		addConfigWatcherIfNotExists: (folder: string, factory: () => api.FileSystemWatcher | null) => void,
```
mweichert, line 439: Is it possible to avoid the `null` here?

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -391,19 +436,30 @@ export const inject = (deps: InjectedDependencies) => {
...
-			return watcher;
+			showOutput(output, OutputLevels.info)(`Adding watcher for directory ${projectDir}.`);
+			try {
+				updateCommandStates(context, output, i18n, projectDir);
+			} catch (error: any) {
+				logAllNestedErrors(error, output);
+			}
+			try {
+				const watcher = vscode.workspace.createFileSystemWatcher(join(projectDir, '.taq/config.json'));
+				// TODO: We should detect the event that VsCode's current Folder is changed and the watcher should be disposed
```
mweichert, line 451: Ali, a convention we are trying to adhere is to always append a `//TODO comment` with a link to an issue in Github. That way we can schedule this work at a later time, and and the work-to-be-done becomes more visible.

mweichert (changes requested):
Good stuff @AlirezaHaghshenas 

I just added a few comments. The most concern I have is the use of `any` and `null` which I have a strong preference to avoid. If you're interested in why let me know and we can discuss.

It will be nice to have the logging in vscode.

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -276,6 +277,45 @@ export const inject = (deps: InjectedDependencies) => {
 		vscode.window.showInformationMessage(msg)
 			.then(_ => Promise.resolve()) as Promise<void>;
 
+	const logAllNestedErrors = (err: TaqVsxError | TaqError | Error | any, output: Output) => {
+		try {
+			if (!err) {
+				return;
+			}
+			if (!shouldOutput(OutputLevels.error, output.logLevel)) {
```
jchenche, line 285: We can do `if (!err || !shouldOutput(OutputLevels.error, output.logLevel)) return;` since the 2 conditions lead to the same execution.

AlirezaHaghshenas: Thanks for the suggestion. The main reason I kept the two separate is to keep it clear for the reader why we are returning. 

- For line 282, we return because we have reached the end of nested errors, this is the break from recursive calls.
- For line 285, we return because the configured level filters out this error.

So while both just `return`, one prevents the output from showing, and another short circuits because there is nothing to output.

I do not insist, by the way, to make such distinctions in Taqueria's code.

taqueria-vscode-extension/src/lib/helpers.ts
AlirezaHaghshenas: It is, but then we have to know all the umbrella types that cover all possible errors. In a language like java, with propagated errors, you know it from the function declaration. In js/ts, you don't know that.

Do you have any workarounds on your mind?

taqueria-vscode-extension/src/lib/helpers.ts
AlirezaHaghshenas: The problem with that is, I can create many such TODOs for a simple task.
A lot of those will just be cleared as we get better insight into vscode, but making sure they are all fine can slow down initial feature implementations even further.

For instance, my current understanding is that this TODO is fine to be deleted, as when you change the folder, the vscode refreshes all the extensions. I checked that and could never see multiple watchers.

If I create issues for all such things, I can pollute the issue tracker with mostly useless issues.

If you still insist, I can do that.

jchenche (approved):

taqueria-vscode-extension/src/lib/helpers.ts
hu3man: I'll second that @AlirezaHaghshenas!

hu3man (approved):
These changes were successful in deployment

mweichert (approved):

AlirezaHaghshenas:
These were merged into another PR and then merged into main, closing.

PR #946, opened 2022-07-05
Fixes #915, adds  template construct

### LINKED ISSUE #915: Plugin Templates
As a plugin author, I would like to expose templates that the end-developer can instantiate to avoid having to write boilerplate

As a plugin author, I'd like the ability to expose named templates, which can thought of as programmable content. A plugin author would compose a template much the same way in which a task is - a template would have a name, command, positional arguments, options, and a handler.

For instance, the ligo plugin could expose a template called _contract_ which could be instantiated using `taq create contract`. This would generate a new ligo template with some boilerplate so that the contract can be compiled and originated.

Yes, it makes it easy for the end-developer to get started with writing contracts.

This functionality is currently in the 572-jest-test-stubs branch which needs to be segmented into its own branch. This issue is required for #907.

A template is defined in a template in much the same way as a plugin author defines a task:
```ts
Template.create({
			template: 'text',
			command: 'text <filename>',
			description: 'Creates a textfile artifact using shell command',
			positionals: [
				PositionalArg.create({
					placeholder: 'filename',
					type: 'string',
					required: true,
					description: 'The name of the filename to generate',
				}),
			],
			options: [
				Option.create({
					shortFlag: 'g',
					flag: 'greeting',
					type: 'string',
					description: 'Greeting to include in JSON file',
				}),
			],
			handler: (requestArgs: RequestArgs.t) => { ... }
            }
})
```

The handler function of a template has the exact same signature as the `proxy` method of a plugin.

### Code of Conduct

mweichert:
Finished:
- [x] Plugin authors can define templates
- [x] Templates are exposed via the `create` task.
- [x] Add support for running template handler, either via proxy or shell
- [x] Add integration tests using mock plugin

Todo:
- [x] Fix failing test in e2e/taqueria-plugin-jest.spec.ts

Deferred:
- Tests failing: e2e/taqueria-plugin-contract-types.spec.ts
- Tests failing: e2e/taqueria-plugin-flextesa.spec.ts


jchenche (approved):
Left 2 comments but other than that, everything looks good!

PR #951, opened 2022-07-06
fix: changed partition's config file to use relative path and updated e2e test accordingly

hui-an-yang:


taqueria-plugin-jest/proxy.ts
```diff
@@ -70,8 +70,8 @@ const createPartition = async (partitionDir: SanitizedAbsPath.t, projectDir: San
 					writeFile(
 						partitionCfgAbsPath,
 						toPartitionCfg(
-							getPartitionAbspath(partitionDir),
-							getRootConfigAbspath(projectDir),
+							SanitizedPath.create(relative(partitionDir, partitionDir)),
```
jchenche, line 73: Why are there two identical `partitionDir` as arguments for `relative()`?

hui-an-yang: It's the relative path from `partitionDir` to itself

jchenche (approved):
LGTM!

alexzbusko (approved):

PR #966, opened 2022-07-08
Refactor Scaffold Task

hu3man:
This PR makes 2 small changes to the scaffold task:
- It removes the `quickstart.md` file injected by `taq init`
- It changes the default scaffold to be `hello-tacos`

### DIFF
--- a/cli.ts
+++ b/cli.ts
@@ -234,12 +234,12 @@ const initCLI = (env: EnvVars, args: DenoArgs, i18n: i18n.t) => {
 					.positional('scaffoldUrl', {
 						describe: i18n.__('scaffoldUrlDesc'),
 						type: 'string',
-						default: 'https://github.com/ecadlabs/taqueria-scaffold-quickstart.git',
+						default: 'https://github.com/ecadlabs/taqueria-scaffold-taco-shop.git',
 					})
 					.positional('scaffoldProjectDir', {
 						type: 'string',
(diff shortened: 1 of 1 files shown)


russellmorton (approved):

jchenche (approved):

cli.ts
```diff
@@ -415,12 +415,14 @@ const scaffoldProject = (i18n: i18n.t) =>
 			await eager(rm(gitDir));
 			log('    ✓ Remove Git directory');
 
-			log('    ✓ Scan for plugins');
-
-			await eager(exec('npm run setup 2>&1 > /dev/null', {}, false, destDir));
-			log('    ✓ Install dependencies');
+			await eager(exec('npm install 2>&1 > /dev/null', {}, false, destDir));
```
mweichert, line 418: @hu3man should we just suppress the stdout? Otherwise if something fails due to a network error than it won't be apparent to the end-developer or the Taqueria team when troubleshooting. Thoughts?

mweichert (commented):
Good stuff. I did leave one question/comment before approving.

cli.ts
hu3man: That's probably the better option @mweichert 

That would just be doing the following right?
```
'npm run setup > /dev/null`
```

mweichert: @hu3man yep, I think that would be better.

PR #972, opened 2022-07-11
implement a GUI

### LINKED ISSUE #965: Spike: VS Code Extension GUI
This ticket will be used to track the initial implementation of UI elements for the Taqueria VS Code Extension.

### Elements to cover in the spike
- [ ] Taqueria icon in the activity bar
- [ ] Once the Taqueria icon is selected in the activity bar, UI element in the Primary Sidebar showing
   - [ ] What plugins are installed (later we can add a command to uninstall them)
   - [ ] What plugins are available to be installed (and an install command on them)
   - [ ] What sandboxes are configured
   - [ ] What environments are configured
   - [ ] What networks are configured
   - [ ] Accounts
   - [ ] Additional information in the config

AlirezaHaghshenas:
closes #965


jchenche (approved):
LGTM!

.vscode/launch.json
```diff
@@ -13,15 +13,16 @@
...
+				"InstallDevPlugins": "false"
 			}
 		},
 		{
 			"request": "launch",
 			"name": "Launch CLI",
 			"type": "pwa-node",
 			"program": "${workspaceFolder}/index.ts",
-			"cwd": "/Users/jimmychenchen/Desktop/bugbug",
+			"cwd": "${workspaceFolder}",
```
jchenche, line 25: I don't think it points to the right folder, because now it points to the Taqueria repo. I think we should leave it `/Users/jimmychenchen/Desktop/bugbug` and maybe and a comment that say: "modify this to point to your taqueria project" or something.

jchenche (approved):

taqueria-vscode-extension/src/lib/pure.ts
```diff
@@ -283,7 +251,17 @@ export const isWindows = () => process.platform.includes('win') && !process.plat
 
 export const findTaqBinary = (i18n: i18n, showOutput: OutputFunction): LikeAPromise<string, TaqVsxError> =>
 	execCmd('which taq', showOutput)
-		.then(path => path.trim())
+		.then(result => {
+			if (result.executionError) {
+				return Promise.reject({
+					code: 'E_TAQ_NOT_FOUND',
+					msg: 'Could not find taq in your path.',
```
hu3man, line 258: Trailing commas can be omitted

taqueria-vscode-extension/src/lib/gui/SandboxesDataProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+		if (!this.workspaceRoot) {
+			pathToDir = null;
+			config = null;
+		} else {
+			try {
+				pathToDir = await Util.makeDir(this.workspaceRoot, this.helper.i18n);
+				config = await Util.TaqifiedDir.create(pathToDir, this.helper.i18n);
+			} catch (e: any) {
+				config = null;
+				await this.helper.notifyAndLogError('Error while loading config, sandboxes list will fail to populate', e);
```
hu3man, line 27: At some point we should refactor this to use i18n but this is fine for now!

hu3man (approved):
Great work on this @AlirezaHaghshenas! It's a big step forward ux wise

.vscode/launch.json
jchenche: `cwd` is supposed to point to the Taqueria project that you're calling `taq` from. That's the right place to point to, which depends on the developer debugging. It's not ideal for now because developers always have to change it to their own folder, but idk how to parametrize this.

GImbrailo (approved):

.vscode/launch.json
AlirezaHaghshenas: But for me, I'm always calling from the same project that I work on.
If we keep this as `/Users/jimmychenchen/Desktop/bugbug`, it will fail for most users.

taqueria-vscode-extension/src/lib/gui/SandboxesDataProvider.ts
AlirezaHaghshenas: Agree. I should keep i18n in mind for next refactors.

.vscode/launch.json
jchenche: Hmm that's ok but maybe make a comment that says "modify to point to where you're invoking taq. e.g. in a Taqueria project"

AlirezaHaghshenas: OK. Give me a moment

jchenche: Reason I tend to wanna leave it like this is that it's more indicative of "point to where your taq project is"

PR #982, opened 2022-07-13
chore: add deno-lock and deno package caching

GImbrailo:
This PR adds the `deno-lock.json` file which will be used to invalidate the `deno` package cache, also part of this PR, when there are any dependency changes.


BGMurphy (approved):

mweichert (commented):
Great for doing this Gino!

I couldn't find a change that ensures that the lock file is used when building the taqueria-cli.

I think we need to add `--lock ./deno-lock.json` to all places where we call `deno compile` or `deno run`.

.github/workflows/main.yml
```diff
@@ -534,4 +559,4 @@ jobs:
         with:
           files: |
             ./release/*
-          draft: true
+          draft: true
```
mweichert, line 562: Do we need a newline here for YAML to be valid?

hu3man (approved):

mweichert (approved):

Issue #324, opened 2022-02-16
Basic contract metadata support


russellmorton:
As a smart contract developer, I want to create, edit and delete and publish *basic* contract metadata to my project

Metadata should be in line with TZIP16 - https://tzip.tezosagora.org/proposal/tzip-16/

Consider:
- TZIP20 - https://tzip.tezosagora.org/proposal/tzip-20/
- TZIP21 - https://tzip.tezosagora.org/proposal/tzip-21/
- Taquito TZIP-16 Contract Metadata and Views - https://tezostaquito.io/docs/metadata-tzip16/
- Taquito TZIP-12 Token Metadata - https://tezostaquito.io/docs/tzip12/

## Creation / Generation

- Global metadata
 - Collected when the user first runs taqueria
 - Persisted to home directory (like opt-in data)
   - Data collected
    - Author (name and e-mail)
    - License
- Project metadata
  - Collected when the user first runs taq init or taq scaffold
  - Persisted to config.json file
  - Data collected
    - Name
    - Description
    - Homepage
    - Authors
    - License (override)
- Contract Metadata:
  - Collected when taq register [contract] is executed
  - Persisted to artfiacts/[contractFilename].json
  - Data:
    - Name (override)
    - Description (override)
    - Version (defaults to md5sum)
    - Homepage (override)
    - License (override)
    - Authors (override)
    - Interfaces
    - If interfaces includes TZIP-12:
      - Name
      - Symbol
      - Decimals

## Validation _Deferred_
Using TZComet
Using Taquito
Compiler validation
Potentially a separate plugin

## Publishing
- Publish via IPFS
- Future consideration:
  - Internal storage
  - HTTP

## Implementation Considerations
- Need to implement hooks
- Collecting global and project-level metadata is done in the CLI (What happens with VS Code? @mweichert)
- Develop a metadata plugin which adds a template called 'metadata'. The user would create an instance of that template just like any other template - `taq create metadata [contractName]`. If invoked with no additional arguments, it would prompt you for each field listed above, however, you could bypass certain fields by passing values as flags. E.g.: `taq create metadata hello.mligo --name "Hello World"`
- Using hooks, we call `taq create metadata` when `taq register [contract]` is called, whether that be directly or indirectly.
- #550 for npm init on taq init
- Implement extended JSON file with commented guidance around contract metadata
- Bigmap instantiation

**USE CASE**
Generate JSON -> Publish to IPFS -> Add hash to contract

[...]

PR #985, opened 2022-07-14
Metadata plugin

ricklove:
Issue: https://github.com/ecadlabs/taqueria/issues/324

- Generate a metadata file for a contract
  - task command `create metadata [contractName]` conflicts with the template system
  - so used `generate metadata  [contractName]` instead
- Prompt the user for standard metadata fields, providing default values.
- If metadata already exists for a contract, it will use existing values as the default prompt values.
- If contractName is not provided, it will prompt for contractName allowing the selection of a taq registered contracts.
- If this is a new contract metadata, it will find other contracts metadata and use those for the default values
  - for common fields like authors, license, homepage
  - this eliminates an extra step to define/register project metadata
- e2e tests are included


ricklove:
Included in #1045

PR #997, opened 2022-07-15
Fix problem with e2e test failing when asserting msg for no contracts

mweichert:
Also, fix deno lock hook.


GImbrailo (approved):

mweichert:
This was merged into #1001

PR #1031, opened 2022-07-21
Return JSON response from CLI

### LINKED ISSUE #1002: CLI should return JSON response instead of printing ASCII table
#987 describes the whole new mechanism of communicating between taqueria core and TVsCE.

However, for now we don't need such complicated refactor.
A temporary solution is to just return JSON string when we query for some complex object or returning an array.

MichalLytek:
Closes #1002 🔒


tests/e2e/taqueria-plugin-flextesa.spec.ts
```diff
@@ -32,6 +32,39 @@ describe('E2E Testing for taqueria typechecker and simulator tasks of the tezos-
...
+					account: 'jane',
+					balance: '3000 ꜩ',
+					address: 'tz1aHUAC4oviwJuZF1EvVSvFz7cu9KMNYBph',
+				},
+				{
+					account: 'joe',
+					balance: '3000 ꜩ',
+					address: 'tz1MVGjgD1YtAPwohsSfk8i3ZiT1yEGM2YXB',
+				},
+			]) + '\n',
```
MichalLytek, line 64: @mweichert Do we have `tests` explicitly excluded from the auto formatting or something is wrong with the config? 😕

jchenche (approved):
LGTM!

tests/e2e/taqueria-plugin-flextesa.spec.ts
mweichert: We do explicitly disable the formatting of tests data.

We were having issues with our strings being reformatted here:
~/Projects/taqueria/tests/e2e/data/help-contents.ts

PR #1043, opened 2022-07-26
Adjusted pre-commit hook to only generate Deno lock file CLI source has changed

mweichert:
We generate a deno.lock file, which is synonymous with package-lock.json for Node, in a pre-commit hook.

However, if someone changes a file in the website, SDK, or a plugin, then we don't need to re-generate the lock file.

This PR adjusts things so that the deno.lock file is only generated when a file relevant to the CLI is modified.


alexzbusko (approved):
LGTM

GImbrailo (approved):
LGTM

PR #1044, opened 2022-07-26
Disable publish step for contract-types plugin

mweichert:
The jest plugin lists the '@taqueria/plugin-contract-types' as a dependency, along with the `@taqueria/node-sdk`. As such, I'll be publishing these manually for the release.

To avoid problems when triggering the release workflow, I'm disabling the publishing of the contract types plugin for now.


alexzbusko (approved):

GImbrailo (approved):

GImbrailo:
FYI the CloudFlare build is failing here because we already have v10 of the sdk in the `package.json` and it does not exist yet. I don't think this should prevent merging

PR #1045, opened 2022-07-26
Metadata Plugin - Contract & Project Metadata

ricklove:
Issue: https://github.com/ecadlabs/taqueria/issues/324

- Generate a metadata file for a contract
  - task command `create metadata [contractName]` conflicts with the template system
  - so used `generate metadata  [contractName]` instead
- Prompt the user for standard metadata fields, providing default values.
- If metadata already exists for a contract, it will use existing values as the default prompt values.
- If contractName is not provided, it will prompt for contractName allowing the selection of a taq registered contracts.
- If this is a new contract metadata, it will find other contracts metadata and use those for the default values
  - for common fields like authors, license, homepage
  - this eliminates an extra step to define/register project metadata, but project metadata will be used as a priority if available
- Project metadata
  - Add the project metadata to .taq config.json
  - Add the project metadata schema
- Summary of priority for **default values**:
  - 1st Priority: Previous Values
  - 2nd Priority: Project Values
  - 3rd Priority: Other Contract Values
  - 4th Priority: Hard-coded defaults 
- e2e tests are included


taqueria-protocol/Config.ts
```diff
@@ -101,6 +102,7 @@ export const internalSchema = commonSchema.extend({
 	environment: environmentMap,
 	accounts: accountsMap,
 	contracts: z.record(Contract.schemas.schema).optional(),
+	metadata: MetadataConfig.schemas.schema.optional(),
```
ricklove, line 105: Is this correct for a flat optional object that should be at the root of the config.json?

```
{
  ....,
  metadata: {
    name: 'Name'
    ...,
  },
}
```

taqueria-plugin-metadata/index.ts
```diff
@@ -0,0 +1,34 @@
...
+					placeholder: 'contractName',
+					description: 'Which contract?',
+					type: 'string',
+				}),
+			],
+			encoding: 'none',
+		}),
+		Task.create({
+			task: 'project-metadata',
+			command: 'generate-project-metadata',
```
ricklove, line 26: Had to use `generate-project-metadata` to avoid a taq bug when multiple commands have the same first word

ricklove: To reproduce:
- Find replace `generate-project-metadata` with `generate project-metadata` (here and in test)
- Rebuild this project
- Run e2e test for this project
- **Result:** All tests will now fail

ricklove:
This should now be ready

taqueria-plugin-metadata/index.ts
ricklove: Reproduced here: #1147

jchenche (approved):

hu3man (approved):

GImbrailo (approved):

sinapsist (approved):
Verified by QA

Issue #1019, opened 2022-07-19
Separate contracts and Artifacts


AlirezaHaghshenas:
### What happened?

There is no natural way to handle SmartPy artifacts. 

May need a separate list for commands on artifacts.

russellmorton:
Deferring to Milestone 3, SmartPy focus.

PR #1058, opened 2022-07-29
Display separate artifacts tree view with originate button

MichalLytek:
Implements #1019

> Currently, we have one TreeView that shows contracts.
But some of the commands that should be run on artifacts are enabled on that view.
So if a user runs an artifact command (like originate) on a contract, the TVsCE has a logic that decides which artifact to run this command on.
This logic is super simple now, and fails in many cases: it replaces the file extension of the contract with .tz to get the artifact file name. This does not work for smartpy.
So a better option would be to have a better distinction of contracts and artifacts. So the user can run the right commands on each.

image

### DIFF
--- a/taqueria-vscode-extension/src/lib/gui/ArtifactsDataProvider.ts
+++ b/taqueria-vscode-extension/src/lib/gui/ArtifactsDataProvider.ts
@@ -1,0 +1,50 @@
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { HasFileName, HasRefresh, VsCodeHelper } from '../helpers';
+
+export class ArtifactsDataProvider implements vscode.TreeDataProvider<ArtifactTreeItem>, HasRefresh {
+	constructor(
+		private workspaceRoot: string,
+		private helper: VsCodeHelper,
+	) {}
+
+	getTreeItem(element: ArtifactTreeItem): vscode.TreeItem {
+		return element;
+	}
+
+	async getChildren(element?: ArtifactTreeItem): Promise<ArtifactTreeItem[]> {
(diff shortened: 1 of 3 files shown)


AlirezaHaghshenas (approved):

taqueria-vscode-extension/src/lib/gui/ArtifactsDataProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+		private workspaceRoot: string,
+		private helper: VsCodeHelper,
+	) {}
+
+	getTreeItem(element: ArtifactTreeItem): vscode.TreeItem {
+		return element;
+	}
+
+	async getChildren(element?: ArtifactTreeItem): Promise<ArtifactTreeItem[]> {
+		const artifacts = await vscode.workspace.findFiles('artifacts/*.*', '**/node_modules/**');
```
AlirezaHaghshenas, line 16: This will work well for Ligo and Archetype artifacts (that are single `.tz` files).
Smartpy creates folders with multiple files in it. This will mess up our artifacts view.

I don't know how we should handle this. I guess we should take the folder's name? @russellmorton

jchenche (approved):

taqueria-vscode-extension/src/lib/gui/ArtifactsDataProvider.ts
MichalLytek: I think SmartPY is a separate milestone. We should track this issue in another ticket to make sure we won't forget about it 😉

hu3man (approved):
🚀

PR #1114, opened 2022-08-11
chore: add workflow to delete old workflows

GImbrailo:
This change will add a workflow that will delete old workflow runs that are no longer required. In the past we have had incidents where old workflow runs can cause confusion when trying to debug a specific issue. One such incident happened previously during a release https://ecadlabs.slack.com/archives/C036G3SVA4Q/p1656525987109809.

Here is an example of some workflows that no longer exist
image

There were a lot more than those listed in the screenshot that were manually deleted. 

Unfortunately we can not see this workflow in action until it is merged into main. I tested it out successfully in the `taqueria-github-action` repo if anyone wants to see how it works before merge.

https://github.com/ecadlabs/taqueria-github-action/runs/7774502009?check_suite_focus=true

### DIFF
--- a/.github/workflows/workflow-pruning.yml
+++ b/.github/workflows/workflow-pruning.yml
@@ -1,0 +1,59 @@
+name: Prune workflow runs
+on:
+  schedule:
+    - cron: '0 0 * * 0'
+  workflow_dispatch:
+    inputs:
+      days:
+        description: 'Number of days to keep'
+        required: true
+        default: 30
+      minimum_runs:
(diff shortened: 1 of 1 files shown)


danielelisi:
It's unclear why this task is required; could you add some context?

GImbrailo:
Thanks @danielelisi, I have updated the PR description

danielelisi (approved):

alexzbusko (approved):

PR #1119, opened 2022-08-13
Change format of storage so that the value is a filename, which contains the actual storage value, in Michelson expression

### LINKED ISSUE #963: [Feature]: have taquito plugin use another representation of storage value

Have taquito plugin use another representation of storage value.

Currently, it's using plain "JSON", but it's having some issues because pure JSON is not enough to represent all Michelson types.

Solution for now is to change the format for specifying storage where users will not specify a value directly in the config file, and instead will specify a file name that contains the actual storage, as a Michelson expression.

### Code of Conduct

jchenche:
## AC
- [x] Implementation
- [x] Documentation
- [x] Testing

Fixes #963 

@hu3man We may need to update to documentations to reflect the fact that we are now mapping contract file name to storage file name in config file.

@mweichert The change log for the next release may need to mention that this is not a backwards compatible change.

@alexzbusko We will need to refactor our e2e tests as a result of this change.


taqueria-sdk/index.ts
```diff
@@ -387,14 +387,20 @@ export const getSandboxAccountConfig = (parsedArgs: RequestArgs.t) =>
...
-	(contractFilename: string) => {
-		const env = getCurrentEnvironmentConfig(parsedArgs);
-
-		return env
-			? env.storage && env.storage[contractFilename]
-			: undefined;
-	};
+export const getInitialStorage = async (parsedArgs: RequestArgs.t, contractFilename: string) => {
+	const env = getCurrentEnvironmentConfig(parsedArgs);
+	if (env && env.storage && env.storage[contractFilename]) {
```
AlirezaHaghshenas, line 392: Not a big deal, just coding style preference:

if we instead do:

```ts
if (!env || !env.storage || !env.storage[contractFilename]) {
  return undefined;
}
```
Then we can reduce the nesting.
Furthermore, we can simplify it to: `if (!env?.storage?.[contractFilename])`

AlirezaHaghshenas (approved):

taqueria-sdk/index.ts
jchenche: hmm I'd usually agree but the nesting is so shallow in this case 😅

taqueria-plugin-taquito/originate.ts
```diff
@@ -49,29 +49,29 @@ const addOrigination = (parsedArgs: Opts, batch: Promise<WalletOperationBatch>)
...
-		(retval, filename) => {
-			const storage = getInitialStorage(parsedArgs)(filename);
+		async (retval, filename) => {
+			const storage = await getInitialStorage(parsedArgs, filename);
 			if (storage === undefined || storage === null) {
 				sendErr(
 					`Michelson artifact ${filename} has no initial storage specified for the target environment.\nStorage is expected to be specified in .taq/config.json at JSON path: environment.${
 						getCurrentEnvironment(parsedArgs)
-					}.storage."${filename}"\n`,
+					}.storage["${filename}]\nThe value of the above JSON key should be the name of the file that contains the actual value of the storage, as a Michelson expression.\n`,
```
alexzbusko, line 68: Missing the closing double quote here for the filename

alexzbusko: Also should this message mention the pathing requirements for the file?

jchenche: Thanks!

hu3man:
@jchenche, as part of the recent role changes, documentation will now need to be created and updated by the developer (with assistance if required)

hu3man:
@alexzbusko will remove the failing test and create a dev-task to implement that again in the future

alexzbusko (approved):

PR #1142, opened 2022-08-18
Verify default method is generated (Add test)

ricklove:
- Run test that checks default method is created as expected:

- Example contract:
  - https://github.com/ecadlabs/taqueria/blob/1542cda9768ef90ffc21888dfa80a541f7701e6f/taqueria-plugin-contract-types/example/contracts/example-contract-10.tz#L1
- Result type with default:
  - https://github.com/ecadlabs/taqueria/blob/1542cda9768ef90ffc21888dfa80a541f7701e6f/taqueria-plugin-contract-types/example/types-file/example-contract-10.types.ts#L9


hu3man (approved):
Thanks for hopping on this so quickly @ricklove!

LGTM

jchenche (approved):

ricklove:
This builds on tests added here in the sub dir fix, so it would be good to wait for that to be approved before merging both:

https://github.com/ecadlabs/taqueria/pull/1140

PR #1174, opened 2022-08-23
Create tests to test VS Code command palette

sinapsist:
Initial tests to verify that the command palette can trigger taqueria plugins


hu3man:
@sinapsist, I see there are some tests failing in the pipeline for this PR. Are you able to investigate?

PR #1309, opened 2022-10-03
Vs code cli testing

sinapsist:
# 🏗️ PR ➾

Fixes # (issue)

## 🪂 Pre-Merge Checklist

- [X] 🐬 I have commented my code, particularly in hard-to-understand areas
- [X] 🦀 Automated tests have been written and added to this PR
- [X] 🤿 Corresponding changes have been made to all documentation
- [X] 🔱 The test plan has been implemented and verified by an SDET
- [X] 🏄‍♂️ Myself and others have built this branch and done manual testing on the change
- [X] ⛵ A summary describing the change has been written for release notes

## 🪁 Description

Please include a summary of the change and add the issue number this is related to above. Please also include relevant motivation and context for the change
(e.g. what was the existing behavior, why did it break, etc.)

## 🎢 Test Plan

Please describe the testing strategy for this change

### 🛸 Type of Change

- [ ] 🐟 Minor change (non-breaking change of very limited scope)
- [ ] 🦑 Bug fix (non-breaking change which fixes an issue)
- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🌊 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🐳 Major refactor (breaking changes and significant impact to the codebase)


taqueria-vscode-extension/src/test/suite/extension.test.ts
```diff
@@ -5,102 +5,112 @@ import * as vscode from 'vscode';
...
+	mockedReturnValues.openDialogMockIndex++;
+	return mockedValue;
+}
+
+let choosePlugin: string = '@taqueria/plugin-smartpy';
+
+const mockedMethods = {
+	'window.showInformationMessage': (msg: string) => Promise.resolve(console.log(msg)),
+	'window.showQuickPick': () => Promise.resolve([choosePlugin]),
+	'window.showOpenDialog': () => {
```
AlirezaHaghshenas, line 51: Not important now, but `window.showOpenDialog` is not backed-up in `originalMethods` in line 18-21.

taqueria-vscode-extension/src/test/suite/utils/utils.ts
```diff
@@ -1,3 +1,27 @@
+import retry from 'async-retry';
+import fsPromises from 'fs/promises';
+
 export function sleep(ms: number): Promise<void> {
 	return new Promise<void>(resolve => setTimeout(resolve, ms));
 }
+
+export async function checkFolderExistsWithTimeout(filePath: string) {
```
AlirezaHaghshenas, line 8: We already have another copy of this function in `taqueria-vscode-extension/src/test/suite/utils/utils.ts`

package.json
```diff
@@ -16,6 +16,7 @@
 		"serve-website": "cd website && npm run serve",
 		"watch": "npx chokidar \"*.ts\" \"taqueria-utils/**/*.ts\" \"taqueria-protocol/**/*.ts\" -c \"npm run build || npx play-sound beep.mp3\" --initial",
 		"build-and-watch-vscode": "npx chokidar \"*.ts\" \"taqueria-vscode-extension/**/*.ts\" \"taqueria-protocol/**/*.ts\" -c \"npm run compile -w taqueria-vscode-extension || npx play-sound beep.mp3\" --initial",
+		"pretest": "mkdir -p ~/TVsCE_e2e/vscode-taq-test-project && npm run compile -w taqueria-vscode-extension",
```
alexzbusko, line 19: Is this command needed since it seems like we are creating the directory for tests when the tests are run?

taqueria-vscode-extension/src/lib/gui/ObservableConfig.ts
```diff
@@ -1,3 +1,4 @@
+import { execSync } from 'child_process';
```
alexzbusko, line 1: This import doesn't look like it's used in this file

taqueria-vscode-extension/src/test/suite/extension.test.ts
```diff
@@ -5,102 +5,112 @@ import * as vscode from 'vscode';
...
+		await taqueriaExtension.activate(context, { vscode: vscodeMock });
 
 		vscode.window.showInformationMessage('Start all tests.');
 	});
 
-	it('Verify that Taqueria Initiate will init new taquifed  project ', async () => {
+	it('Verify that Taqueria commands present in the command pallet list ', async () => {
 		await vscode.commands.getCommands(true).then(allCommands => {
 			const taqCommands = allCommands.filter(command => command.toLowerCase().includes('taq'));
 			assert.notEqual(taqCommands, undefined);
```
alexzbusko, line 71: If we are checking that the commands are there I think we can use this test as a list containing all those commands so we know if anything changes rather than just checking that the commands are not undefined

taqueria-vscode-extension/src/test/suite/extension.test.ts
```diff
@@ -5,102 +5,112 @@ import * as vscode from 'vscode';
...
-		// Example how to use sleep to stop VS Code for a moment
-		// await sleep(60000);
 	});
 
-	// TODO: https://github.com/ecadlabs/taqueria/issues/605
-	it.skip('Verify that VS Code command Taqueria Scaffold will scaffold a project', async () => {
-		// Open taquified project in workspace under testProjectDestination
+	it('Verify that VS Code command Taqueria Compile Ligo will compile Ligo contract', async () => {
+		// It creates another process
+		// https://stackoverflow.com/questions/51385812/is-there-a-way-to-open-a-workspace-from-an-extension-in-vs-code
```
alexzbusko, line 77: I think a lot of this commended out code can be removed since it doesn't seem to be relevant

taqueria-vscode-extension/src/test/suite/extension.test.ts
```diff
@@ -5,102 +5,112 @@ import * as vscode from 'vscode';
...
+		choosePlugin = '@taqueria/plugin-ligo';
 
-	// TODO: https://github.com/ecadlabs/taqueria/issues/645
-	it.skip('Verify that VS Code command Taqueria Compile Ligo will compile Ligo contract', async () => {
-		// Install plugin
-		await vscode.commands.executeCommand('taqueria.install @taqueria/plugin-ligo');
+		await vscodeMock.commands.executeCommand('taqueria.install');
 
-		await vscode.commands.executeCommand('taqueria.compile_ligo');
+		choosePlugin = '@taqueria/plugin-flextesa';
```
alexzbusko, line 92: Why are you now installing the flextesa plugin? I don't think this is needed for this test

tests/package.json
```diff
@@ -9,6 +9,7 @@
 		"test:unit:deno": "TEST=true deno test unit/ --allow-all --unstable --import-map ../import_map.json",
 		"test:unit:jest": "TEST=true jest --config jest.config.unit.ts --coverage --no-cache",
 		"test:unit:deno:debug": "TEST=true deno test unit/ --allow-all --unstable --inspect-brk",
+		"test:unit:protocol": "TEST=true deno test unit/taqueria-protocol --allow-all --unstable --inspect-brk",
```
alexzbusko, line 12: Is this really needed for the VSCode tests?

taqueria-vscode-extension/src/test/suite/extension.test.ts
```diff
@@ -5,102 +5,112 @@ import * as vscode from 'vscode';
...
+		await fse.copyFile(ligoContractFileSource, ligoContractFileDestination);
 
-	// TODO: https://github.com/ecadlabs/taqueria/issues/688
-	it.skip('Verify that VS Code command Taqueria Deploy Contract will deploy contract to a test network', async () => {
-		// Install plugin
-		await vscode.commands.executeCommand('taqueria.install @taqueria/plugin-smartpy');
+		await vscodeMock.commands.executeCommand('taqueria.compile_pick_file');
 
-		await vscode.commands.executeCommand('taqueria.deploy');
+		assert.doesNotThrow(() => {
```
alexzbusko, line 103: If we are just testing that the expected file matches the file output I don't think we need to wrap this all in an assertion that no error happens. If an error occurs the assertion on the files should tell us what happened

taqueria-vscode-extension/src/test/suite/extension.test.ts
```diff
@@ -5,102 +5,112 @@ import * as vscode from 'vscode';
...
+		|| mockedReturnValues.openDialogMockIndex >= mockedReturnValues.openDialogMockValues.length
+	) {
+		throw new Error('We have already used all mocked values.');
+	}
+	const mockedValue = mockedReturnValues.openDialogMockValues[mockedReturnValues.openDialogMockIndex];
+	mockedReturnValues.openDialogMockIndex++;
+	return mockedValue;
+}
+
+let choosePlugin: string = '@taqueria/plugin-smartpy';
```
alexzbusko, line 46: Should we default to LIGO as a first choice for plugins?

taqueria-vscode-extension/src/test/suite/extension.test.ts
```diff
@@ -5,102 +5,112 @@ import * as vscode from 'vscode';
...
-		// Run taqueria scaffold command
-		await vscode.commands.executeCommand('taqueria.scaffold');
-	});
+		setOpenDialogMocks(`${ligoContractFileDestination}`);
 
-	// TODO: https://github.com/ecadlabs/taqueria/issues/686
-	it.skip('Verify that VS Code command Taqueria Start Sandbox starts flextesa sandbox', async () => {
-		// Install plugin
-		await vscode.commands.executeCommand('taqueria.install @taqueria/plugin-flextesa');
+		await fse.rm(testProjectDestination, { recursive: true });
```
alexzbusko, line 83: Why are you removing the test project destination and then making it within the test? Shouldn't the project destination be made by the run-tests.sh script before the tests are run?

.vscode/launch.json
```diff
@@ -18,34 +19,21 @@
...
-				"--inspect-brk",
-				"--allow-run",
-				"--allow-write",
-				"--allow-read",
-				"--allow-env",
-				"--allow-net",
-				"--import-map",
-				"${workspaceFolder}/import_map.json"
-			],
+			"runtimeExecutable": "${execPath}",
```
alexzbusko, line 25: I'm not sure what `${execPath}` refers to here. What does this mean and what does it do?

alexzbusko (commented):
I have some questions and left some comments @sinapsist. Glad to see this is all coming together

alexzbusko:
One thing that would also help people get this going is something in a README.md file in the vscode tests directory @sinapsist. I don't know how to get these up and running on my machine right now and could use a pointer or two

.vscode/launch.json
sinapsist: Hey @alexzbusko, more details about it here - https://code.visualstudio.com/docs/editor/variables-reference
${execPath} - the path to the running VS Code executable

package.json
sinapsist: Hey @alexzbusko. Yes, we need to create it before we compile the project. And we clean-up and refresh it every test. But if we don't do it before compiling, it won't pick up as a workspace folder.

taqueria-vscode-extension/src/lib/gui/ObservableConfig.ts
sinapsist: Yeah, good point. I removed it.

taqueria-vscode-extension/src/test/suite/extension.test.ts
sinapsist: @alexzbusko I would keep it because it is a reminder of why it cannot be done this way. I spent a lot of time debugging it, and it is nice to have a reminder about it.

taqueria-vscode-extension/src/test/suite/extension.test.ts
sinapsist: You're right, we just tested that multiple plugins could be installed. Removing this step.

taqueria-vscode-extension/src/test/suite/extension.test.ts
sinapsist: @alexzbusko No, we are also testing that file has been created. If it is not created it throws an error. That is why it is wrapped into doesNotThrow assertion here.

tests/package.json
sinapsist: No, I've removed it.

taqueria-vscode-extension/src/test/suite/utils/utils.ts
sinapsist: @AlirezaHaghshenas Yeas, and this is it. Sorry, I am not getting a bit this comment. It is the same file.

taqueria-vscode-extension/src/test/suite/extension.test.ts
sinapsist: Yes, but for each test, we do a clean-up of the folder - rm and mkdir on lines 84,85 are doing it for us. Because we want to have it clean for each test without plugins, etc. But we have to create a reference first (that is what we do in run-tests.sh, otherwise it won't be picked up by VS Code)

taqueria-vscode-extension/src/test/suite/extension.test.ts
sinapsist: I updated test logic to check list of commmands.

alexzbusko (approved):

AlirezaHaghshenas (approved):

GImbrailo (approved):
The workflow changes LGTM

.github/workflows/vscode-extension.yml
```diff
@@ -45,20 +45,35 @@ jobs:
...
       - name: Install Dependencies
         run: npm ci
 
       - name: Compile packages
-        run: npm run build-all
+        run: |
+          echo "$(pwd)" >> $GITHUB_PATH
+          npm run build-all
+
+      - name: Make taq binary globally visible
```
alexzbusko, line 64: If the directory output from `pwd` is the same as the directory output from which taq in this step are the same I think this step can be removed since it will be accessible if the above command works

.github/workflows/vscode-extension.yml
```diff
@@ -45,20 +45,35 @@ jobs:
...
+
+      - name: update run-tests.sh to make it executable
+        run: chmod +x ./taqueria-vscode-extension/bin/run-tests.sh
       
       - name: Compile TS
         run: npm run compile:test --workspace ${{ env.GIT_WORKSPACE }}
       
       - name: Run Unit Tests on Linux
-        run: xvfb-run -a npm test --workspace ${{ env.GIT_WORKSPACE }}
+        run: |
```
alexzbusko, line 75: Doesn't need to be piped since there is only one command

taqueria-vscode-extension/src/test/suite/extension.test.ts
```diff
@@ -1,107 +1,143 @@
 import assert from 'assert';
+import { execSync } from 'child_process';
```
alexzbusko, line 2: This doesn't look like it is getting used in this file

.github/workflows/vscode-extension.yml
sinapsist: Fixed.

.github/workflows/vscode-extension.yml
sinapsist: Fixed.

taqueria-vscode-extension/src/test/suite/extension.test.ts
sinapsist: Fixed.

PR #1310, opened 2022-10-04
refactor(ci): publish workflows

### LINKED ISSUE #1297: 🚧 Dev ➾ Refactor release workflows
### 🚥 Definition of Done

- [x] 🎯 Planning and Design
- [ ] 🏎️ Dev Implementation
- [ ] 🏁 Test Plan Executed
- [ ] 🛩️ Documentation Updated
- [ ] 🏟️ Peer / User Reviewed
- [ ] 🏆 Validated and Accepted

### ✅ To Do

- [ ] Prereleases have the `-rc` or `-beta` suffix
- [ ] NPM publish workflow simplified
- [x] Caching built packages to prevent building packages multiple time in different steps #1310 
- [ ] Nice-to-have: Prerelease from a manual workflow

### 🔭 Scope

TBD

### 🧰 Work Description

In an effort to continually improve our workflows there are a number of changes that we can make. This issue aims to highlight those changes and prioritise them.

## Prerelease suffix
It is best practice for our prereleases to have a suffix in the tag that can quickly inform the user that they are working with a prerelease. We moved away from this to using the odd/even minor version to determine prerelease or full release. We could have both and it would be beneficial to our users

## NPM workflows simplified
There is room to improve how we call the reusable workflow especially in the areas where we call the reusable workflow for each package. This could potentially be done once or twice maximum.

## Caching built packages
We build the taqueria packages multiple times throughout all the workflows, this is a waste of resources and also takes very long. We can build these packages once, cache them, and then use the cache whenever they need to be built again.

## Prerelease from a manual workflow
Previously we discussed being able to kick off a prerelease using a manual workflow and predetermined inputs. This is relatively easy and will help standardise our release process.

### ⚖️ Acceptance Criteria

TBD

### 🎢 Test Plan

TBD

### 🎆 Release Notes

_No response_

### 🦄 Sizing

5

### Code of Conduct

GImbrailo:
# 🏗️ PR ➾ refactor(ci): publish workflows

Fixes #1297

## 🪂 Pre-Merge Checklist

## 🪁 Description
Our workflows for releases currently build the `npm` packages in multiple places which increase the workflow run times. There is also currently an issue with how the workflow checks for a full release or a prerelease, and can be resolved by adding back the `-rc` or `-beta` suffixes to the release versions and doing checks on the suffix. Finally the `build-all` script and the `npm publish` workflows can be improved and simplified using `npm workspaces`

- [x] Use `npm workspaces` to simplify the `npm build` and `npm publish` steps
- [x] Build and cache the npm packages to reduce build times throughout the project
- [ ] ~~Re-add the `-rc` or `-beta` suffix checks~~

### 🛸 Type of Change

- [x] 🐟 Minor change (non-breaking change of very limited scope)


GImbrailo:
@mweichert I am just dealing with a failing test in the VSCode workflow but everything else is ready for review. I am hoping to get your input on this, specifically the changes to the `build-all` script.

jchenche (approved):

hu3man (approved):

BGMurphy (approved):

.github/workflows/vscode-extension.yml
```diff
@@ -52,20 +52,18 @@ jobs:
...
+          path: ./node_modules/**
+          key: cache-node-modules-${{ github.run_id }}
       
       - name: Compile TS
         run: npm run compile:test --workspace ${{ env.GIT_WORKSPACE }}
       
-      - name: Run Unit Tests on Linux
-        run: xvfb-run -a npm test --workspace ${{ env.GIT_WORKSPACE }}
+      # - name: Run Unit Tests on Linux
+      #   run: xvfb-run -a npm test --workspace ${{ env.GIT_WORKSPACE }}
```
danielelisi, line 66: Why are you disabling unit tests?

.github/workflows/main.yml
```diff
@@ -57,37 +58,57 @@ jobs:
...
+
+      - name: Setup Deno
+        uses: denoland/setup-deno@v1
+        with:
+          deno-version: "1.23.4"
+
+      - name: Build all
+        run: |
+          export PATH="$PATH:$(pwd)"
+          npm run build-all
```
danielelisi, line 81: Why are you building the packages here if what you're caching are only the `node_modules` ? Shouldn't just an `npm ci` work here?

.github/workflows/vscode-extension.yml
hu3man: The tests don't work 😔

.github/workflows/main.yml
GImbrailo: This is the initial build that will build all the `node_modules` that will be used later on. Because the cache is unique for each run I need to build all the modules at least once.

bin/build-all.sh
```diff
@@ -11,32 +11,8 @@ if [ "$0" == "./bin/build-all.sh" -a -f index.ts ]; then
...
-    for dir in `ls -1d *plugin*`; do
-        if [ -d $dir ]; then
-            echo ""
-            echo '**********************************************'
-            echo "** Building ${dir}"
-            ./bin/build-plugin.sh $dir
-        fi
-    done
+    echo "** Building packages"
+    npm run build --workspaces --if-present
```
mweichert, line 15: @GImbrailo a couple of thoughts here:

I like the simplicity of the single build command.

At this time, the taqueria CLI is in the root package.json. The root isn't considered a workspace, and thus we'll need to add a `npm run build` as well. 

I've always wanted to adjust the build script to: 
1) build packages in parallel (where possible);
2) show aggregate build success/failure messages post-building. 
3) halt building a package if any dependencies failed to build

I'm mentioning this because I'm not sure of how that will impact this script. Perhaps using a tool like the following would be sufficient, but I haven't had a chance to test it: `https://www.npmjs.com/package/workspaces-run/v/1.0.0`

Lastly, I have a concern about whether things get built in the correct order when not manually controlled. @alexzbusko perhaps you could try this? To do so:
1) Delete all build artifacts so that you have a clean state
2) Run `npm run build --workspaces --if-present`
3) Install the flextesa and taquito plugin
4) Try originating a contract

I'll try running through the same procedure a little later today.

--
Just while I'm thinking of it and to share those plans with you @Gino, there are some other adjustments I'd like to eventually make to the build process. I'll mention these things mostly for context, but if you think now might be a good time to roll some of these adjustments out, then perhaps that's something we could discuss:
- Taqueria CLI files should be moved to their own workspace. This will make editing in VSCode much easier, as we can switch between Deno and TypeScript language servers without having to open packages in separate windows
- Have build artifacts always go to a `dist` or `build` directory
- Have a `npm run clean` task to remove and clean up any built artifacts

package.json
```diff
@@ -58,6 +58,12 @@
 		"with-staged": "^1.0.2"
 	},
 	"workspaces": [
+		"./taqueria-protocol",
```
mweichert, line 61: Why do we need to list each workspace here? Shouldn't the `./taqueria*` cover it?

taqueria-flextesa-manager/package.json
```diff
@@ -1,6 +1,7 @@
 {
 	"name": "@taqueria/flextesa-manager",
 	"version": "0.20.2",
+	"private": true,
```
mweichert, line 4: Nice addition!

mweichert (commented):
Very good work, Gino. Here's a summary of my findings:

- I'd like to get confirmation from @alexzbusko that a specific testing procedure works. Alex, I've pinged you in a comment I left in this PR with some steps for a testing procedure I'd like us to try and verify. I'll be running through the same process later today.

- We're building node modules once per run. We might be able to do better by just building when the hash of the package-lock.json has changed

- In the build-all.sh script, we need a `npm run build` statement as well to build the Taqueria binary. 

- To speed up the workflow even further, I think we could move all build artifacts to their own `dist/build` directory. We could then cache those for each run which should have a dramatic effect. This might be best to go in a future PR

- I've been meaning to evaluate packages that allow us to build our packages in parallel. There's some tools that might be able to help with that, and I'm not sure whether it makes sense to tackle that in this PR or not. Here's the list of tools to evaluate: 1) [ultra-runner](https://www.npmjs.com/package/ultra-runner); 2) [workspaces-run](https://www.npmjs.com/package/workspaces-run/v/1.0.0)

- We'd added quite a few globs to the workspaces config directive in package.json. Are those strictly needed?

- I'm listed some future plans, like moving the taqueria-cli files to its own workspace, among some other plans/ideas. Do you think it makes sense to tackle any of those now, or hold for a future PR?

Again, great stuff here.

bin/build-all.sh
GImbrailo: Thanks @mweichert. When packages will get built in the order in which they are defined in the `workspaces` directive of the `package.json` file. I think you raised this in that section of code but I will mention it here in one place. Previously we had `./taqueria*` and when we use globs `workspaces` does not do it in any specific order. Now we have specified so `workspaces` will build `./taqueria-protocol`, then `./taqueria-sdk` ans so on until it reaches line 67 where we still define `./taqueria*` asn it will build anything else that it has not built in previous steps.

You can find more information on this in the official [workspaces docs](https://docs.npmjs.com/cli/v7/using-npm/workspaces), specifically what is in the screenshot below
image
 
I don't think the steps you gave @alexzbusko will work. The best is to run `npm run build all` as this will run in order:
```
npm install
npm run build --workspaces --if-present
npm run build
npm run build-docker --workspaces --if-present
```

package.json
GImbrailo: Mentioned elsewhere

.github/workflows/main.yml
```diff
@@ -347,19 +239,28 @@ jobs:
       - gather-info-for-e2e-tests
     if: ${{ always() && !(contains(needs.*.result, 'failure')) && github.event_name != 'release' && !(startsWith(github.ref, 'refs/tags/v')) }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
```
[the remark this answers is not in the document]
GImbrailo: We are definitely planning for this when we get the Mac minis running in our infrastructure.

.github/workflows/main.yml
```diff
@@ -57,37 +58,57 @@ jobs:
...
+      - name: Build all
+        run: |
+          export PATH="$PATH:$(pwd)"
+          npm run build-all
+
+      - name: Cache node modules
+        uses: actions/cache@v3
+        with:
+          path: ./node_modules/**
+          key: cache-node-modules-${{ github.run_id }}
```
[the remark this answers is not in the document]
GImbrailo: I was not sure of this. Does the `package-lock.json` file change if I make a change to a file like `taqueria-plugin-ligo/compile.ts` for example?

GImbrailo:
@mweichert WRT the following:
>I've been meaning to evaluate packages that allow us to build our packages in parallel. There's some tools that might be able to help with that, and I'm not sure whether it makes sense to tackle that in this PR or not. Here's the list of tools to evaluate: 1) [ultra-runner](https://www.npmjs.com/package/ultra-runner); 2) [workspaces-run](https://www.npmjs.com/package/workspaces-run/v/1.0.0)

In a separate branch I have been testing out running `lerna` with `workspaces` with great success. `lerna` not only makes these parallel builds very easy but more importantly makes building and publishing our packages much easier both locally and in CI. Introducing `lerna` with `workspaces` and `nx` would have almost no disruption to our team but will introduce a lot of stability to our builds. I have a POC ready to go and would love to share it with you ASAP. I did not want to put all the changes in a single PR though as the changes in this PR are valuable without `lerna`.

bin/build-all.sh
GImbrailo: Also WRT the following:
>At this time, the taqueria CLI is in the root package.json. The root isn't considered a workspace, and thus we'll need to add a npm run build as well.

I did not remove the `npm run build` in the project root so it is not showing up in the changes but we are definitely still running a `build` in the root

danielelisi (approved):

.github/workflows/main.yml
```diff
@@ -57,37 +58,57 @@ jobs:
...
+          cache: 'npm'
+
+      - name: Setup Deno
+        uses: denoland/setup-deno@v1
+        with:
+          deno-version: "1.23.4"
+
+      - name: Build all
+        run: |
+          export PATH="$PATH:$(pwd)"
```
alexzbusko, line 80: I think the preferred way to get things in the path is through appending it to the $GITHUB_PATH env var like they do here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path

alexzbusko (approved):
One little thing but overall I am very excited for these changes

.github/workflows/main.yml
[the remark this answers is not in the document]
mweichert: It does not. The package-lock.json file would get updated whenever you modify a package.json file and run `npm update` (which is what we all should be doing if we modify the package.json file for any package).

I suppose why you ask that, is because the node_modules folder also contains packages we've developed, not just third-party modules. However, these should be installed via symlink and so the symlink should still point to the same place.

When we cache node_modules, are the file types of these symlinks preserved as a link, or are they resolved first and then cached?

.github/workflows/main.yml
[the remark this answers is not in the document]
mweichert: Great!

mweichert (approved):

package.json
mweichert: Thanks!

.github/workflows/main.yml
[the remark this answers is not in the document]
GImbrailo: I think they are resolved and then cached as I am only caching the `node_modules` in the project root and all compiled files in the plugin directories are available in the cache

PR #1384, opened 2022-10-18
1375 tezos client rpc target jcc

jchenche:
### ✅ Required for Merging

- [ ] 🏄‍♂️ Myself and others have built this branch and done manual testing on the change
- [ ] ⛵ Documentation is added/updated for the release version 
- [ ] 🖥 Required changes to the VScE have been made
- [ ] 👷 Required updates to scaffolds have been made

### 🙅 Not-Required

- [ ] 🛑 (drag items from above to this list to remove from requirements)

## 🪁 Description

Please include a summary of the changes to the codebase. Please also include relevant motivation and context for the change
(e.g. what was the existing behavior, why did it break, etc.)

## 🎢 Test Plan

Please describe the testing strategy and plan for this PR

### 🛸 Type of Change

- [ ] 🧹 Chore ➾
- [ ] 🛠️ Fix ➾
- [ ] ✨ Feature ➾
- [ ] 🏗️ Refactor ➾
- [ ] 🚀 Release ➾

### DIFF
--- a/taqueria-plugin-tezos-client/common.ts
+++ b/taqueria-plugin-tezos-client/common.ts
@@ -12,14 +12,15 @@ export interface SimulateOpts extends RequestArgs.ProxyRequestArgs {
 	entrypoint?: string;
 }
 
-// To be used for the main entrypoint to the plugin
+// To be used for the main entrypoint of the plugin
 export type IntersectionOpts = TypeCheckOpts & SimulateOpts;
 
 // To be used for common functions in this file
 type UnionOpts = TypeCheckOpts | SimulateOpts;
 
-export const FLEXTESA_IMAGE = 'oxheadalpha/flextesa:20220510';
-export const GLOBAL_OPTIONS = '--endpoint https://jakartanet.ecadinfra.com';
+export const FLEXTESA_IMAGE = 'oxheadalpha/flextesa:rc-20220915-arm64';
+export const GLOBAL_OPTIONS =
+	'--endpoint https://ghostnet.ecadinfra.com --protocol PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg';
 
 export const trimTezosClientMenuIfPresent = (msg: string): string => {
 	return msg.replace(/Usage:(.|\n)+/, '');
--- a/taqueria-plugin-tezos-client/common.ts
+++ b/taqueria-plugin-tezos-client/common.ts
@@ -1,0 +1,38 @@
+import { RequestArgs } from '@taqueria/node-sdk/types';
+import { join } from 'path';
+
+export interface TypeCheckOpts extends RequestArgs.ProxyRequestArgs {
+	sourceFile: string;
+}
+
+export interface SimulateOpts extends RequestArgs.ProxyRequestArgs {
+	sourceFile: string;
+	storage?: string;
+	param: string;
+	entrypoint?: string;
+}
+
+// To be used for the main entrypoint to the plugin
+export type IntersectionOpts = TypeCheckOpts & SimulateOpts;
+
+// To be used for common functions in this file
+type UnionOpts = TypeCheckOpts | SimulateOpts;
+
+export const FLEXTESA_IMAGE = 'oxheadalpha/flextesa:20220510';
+export const GLOBAL_OPTIONS = '--endpoint https://jakartanet.ecadinfra.com';
+
+export const trimTezosClientMenuIfPresent = (msg: string): string => {
+	return msg.replace(/Usage:(.|\n)+/, '');
+};
+
+export const getInputFilename = (opts: UnionOpts, sourceFile: string) =>
+	join('/project', opts.config.artifactsDir, sourceFile);
+
+export const getCheckFileExistenceCommand = (parsedArgs: UnionOpts, sourceFile: string): string => {
+	const projectDir = process.env.PROJECT_DIR ?? parsedArgs.projectDir;
+	if (!projectDir) throw `No project directory provided`;
+	const baseCmd = `docker run --rm -v \"${projectDir}\":/project -w /project ${FLEXTESA_IMAGE} ls`;
+	const inputFile = getInputFilename(parsedArgs, sourceFile);
+	const cmd = `${baseCmd} ${inputFile}`;
+	return cmd;
+};
--- a/taqueria-plugin-tezos-client/client.ts
+++ b/taqueria-plugin-tezos-client/client.ts
@@ -1,14 +1,7 @@
 import { sendAsyncErr } from '@taqueria/node-sdk';
-import { RequestArgs } from '@taqueria/node-sdk/types';
+import { IntersectionOpts as Opts } from './common';
 import simulate from './simulate';
 import typecheck from './typecheck';
-
-interface Opts extends RequestArgs.ProxyRequestArgs {
-	sourceFile: string;
-	storage?: string;
-	param: string;
-	entrypoint?: string;
-}
 
 export const client = (parsedArgs: Opts): Promise<void> => {
 	switch (parsedArgs.task) {
--- a/taqueria-plugin-tezos-client/simulate.ts
+++ b/taqueria-plugin-tezos-client/simulate.ts
@@ -7,30 +7,17 @@ import {
 	sendJsonRes,
 	sendWarn,
 } from '@taqueria/node-sdk';
-import { RequestArgs } from '@taqueria/node-sdk/types';
-import { basename, extname, join } from 'path';
-
-interface Opts extends RequestArgs.t {
-	sourceFile: string;
-	storage?: string;
-	param: string;
-	entrypoint?: string;
-}
+import { basename, extname } from 'path';
+import {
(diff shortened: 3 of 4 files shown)


hu3man (approved):

alexzbusko:
@jchenche I was looking through the e2e tests for  tezos-client after this and I am seeing the following error which is causing all the failures on this branch. Can you take a look into why this is happening and a possible solution?

```
=== For hello-tacos.tz ===
WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
standard_init_linux.go:228: exec user process caused: exec format error
```

jchenche:
> @jchenche I was looking through the e2e tests for tezos-client after this and I am seeing the following error which is causing all the failures on this branch. Can you take a look into why this is happening and a possible solution?
> 
> ```
> === For hello-tacos.tz ===
> WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
> standard_init_linux.go:228: exec user process caused: exec format error
> ```

Ahhh I know what's going on. The image I chose is for arm, but we need logic to determine if we should use an image for arm or amd

taqueria-plugin-tezos-client/common.ts
```diff
@@ -0,0 +1,39 @@
...
+	entrypoint?: string;
+}
+
+// To be used for the main entrypoint of the plugin
+export type IntersectionOpts = TypeCheckOpts & SimulateOpts;
+
+// To be used for common functions in this file
+type UnionOpts = TypeCheckOpts | SimulateOpts;
+
+export const FLEXTESA_IMAGE = 'oxheadalpha/flextesa:latest';
```
mweichert, line 21: I recall hearing that this change is temporary, but should we use the `latest` tag here? I think it might be best to fetch the latest version from dockerhub, and use that instead.

taqueria-plugin-tezos-client/common.ts
```diff
@@ -0,0 +1,39 @@
...
+export const GLOBAL_OPTIONS =
```
mweichert, line 22: Again, despite this being temporary, if it's temporary long enough, and a node goes down, then users might run into the same situation and dependent on us releasing a fix. To mitigate against that, perhaps we'd be best to have these things overridable via environment variables. So perhaps we could use:

```
const endpoint = process.env['TAQ_TEZOS_CLIENT_RPC'] ?? 'https://ghostnet.ecadinfra.com'`

const protocol = process.env['TAQ_TEZOS_CLIENT_PROTOCOL'] ??
'PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg'
```

mweichert (changes requested):
Hey Jimmy, looks good overall. I did add a couple of comments and wishes that might help make this a little more flexible/adaptable, which could be useful if our environment/network adjustments take a bit (which I do suspect they will, as we will mostly rollout improvements in iterations).

Additionally, I believe the flextesa plugin has some logic for determining the platform, and using that information to pick the correct image. It probably makes sense of adding that to the SDK so that any other plugin using docker can make use of it.

taqueria-plugin-tezos-client/common.ts
jchenche: Good idea

hu3man: Awesome, great suggestion @mweichert!

AlirezaHaghshenas (approved):

taqueria-plugin-tezos-client/common.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const trimTezosClientMenuIfPresent = (msg: string): string => {
+	return msg.replace(/Usage:(.|\n)+/, '');
+};
+
+export const getInputFilename = (opts: UnionOpts, sourceFile: string) =>
+	join('/project', opts.config.artifactsDir, sourceFile);
+
+export const getCheckFileExistenceCommand = async (parsedArgs: UnionOpts, sourceFile: string): Promise<string> => {
+	const projectDir = process.env.PROJECT_DIR ?? parsedArgs.projectDir;
```
mweichert, line 33: I don't think we should have plugins look for environment variables for core configuration settings. We should have the CLI do that so that the plugin can just use parsedArgs.

taqueria-sdk/index.ts
```diff
@@ -100,6 +101,19 @@ export const getArch = (): LikeAPromise<string, TaqError> => {
 	}
 };
 
+export const getLatestFlextesaImage = async (arch: 'linux/arm64/v8' | 'linux/amd64'): Promise<string> => {
```
mweichert, line 104: Beautiful!  Do you mind adding a test for this to the SDK?

mweichert (changes requested):
Excellent work, Jimmy! Looks awesome. I did leave a comment or two, and then had a couple of requests, such as a test for the getLatestFlextesaImage function.

taqueria-sdk/index.ts
```diff
@@ -36,6 +36,7 @@ import crypto from 'crypto';
 
 // @ts-ignore interop issue. Maybe find a different library later
 import { templateRawSchema } from '@taqueria/protocol/SanitizedArgs';
+import fetch from 'node-fetch';
```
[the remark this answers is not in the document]
jchenche: I found this to be a suitable enough library for our case.

taqueria-sdk/index.ts
jchenche: will do!

taqueria-plugin-tezos-client/common.ts
jchenche: What's wrong with that? A plugin author may do such a thing and it wouldn't be practical for them to wait for us to update the CLI to accommodate this no?

taqueria-sdk/index.ts
```diff
@@ -100,6 +101,19 @@ export const getArch = (): LikeAPromise<string, TaqError> => {
...
+	const tags = data.results;
+	for (const tag of tags) {
+		if (tag.name === 'latest') continue;
+		for (const image of tag.images) {
+			if (arch.includes(image.architecture)) return `oxheadalpha/flextesa:${tag.name}`;
+		}
+	}
+	return Promise.reject(new Error(`No flextesa image found for the platform of this device, which is ${arch}`));
+};
+
```
mweichert, line 116: Jimmy, I see that we're missing logic to ensure that this is the latest image?

jchenche: I don't think that's possible because the latest image changes overtime. And in this version, the `latest` only points to the "latest" for Intel chip.

mweichert: What I had in mind was a method like:
```
function getDockerImageTags(dockerImageName: string, filter: FilterFn): DockerTags[] where DockerTag is {name, arch} 
```

And then:
```
function getLatestDockerImageTag(dockerImageName: string, filter: FilterFn): DockerTag
```

FilterFn would be `(tag: DockerTag) => bool` that allows the caller to filter the list of tokens before returning the result. This might be necessary to distinguish rc/beta tags from stable ones.

jchenche: In this version of flextesa, the M1 version is `rc` and I don't think they have plans to make it `stable`. Maybe they forgot to change the name, so I don't think such distinction is meaningful.

taqueria-plugin-tezos-client/common.ts
mweichert: For first-party products, we have some more control and options about where we do this kind of thing.

I think it would make sense to have the CLI interpret PROJECT_DIR rather than have each plugin potentially behave differently. If PROJECT_DIR environment variable is absolutely needed here, then I think it belongs in the CLI. Otherwise, I'd just remove checking for that environment variable now and we can add a separate task/PR for that effort. What do you think?

taqueria-sdk/index.ts
```diff
@@ -100,6 +101,9 @@ export const getArch = (): LikeAPromise<string, TaqError> => {
 	}
 };
 
+export const getFlextesaImage = (arch: 'linux/arm64/v8' | 'linux/amd64'): string =>
```
mweichert, line 104: I'm not sure whether this should be in the SDK or not. Seems like a SoC violation, no?

jchenche: Should I put it in the plugin instead?

jchenche: Though I think it's general enough to be put here 🤔

taqueria-sdk/index.ts
jchenche: This is an outdated change btw

taqueria-plugin-tezos-client/common.ts
mweichert: Sounds good!

taqueria-sdk/index.ts
mweichert: I could go either way on this, so I'll leave it up to you. I just wanted to probe your thoughts.

PR #1445, opened 2022-11-03
✨ Feature ➾ Allow docker images to be overwritten via environment variable (#1425)

### LINKED ISSUE #1425: 🚧 Dev ➾ Provide a means of overriding all docker images
### 🔗 Epic|Feature Link

_No response_

### 🚥 Definition of Done

- [x] 🎯 Planning and Design
- [ ] 🏎️ Dev Implementation Complete
- [ ] ♾️ PR Merged
- [ ] 🏆 Product Acceptance

### ✅ To Do

- [ ] #1508
- [ ] Determine whether each plugin that uses a docker image should provide a "show-image-name" task.

### ⚖️ Acceptance Criteria

✨ Functional: 
- [ ] Add a function to the Taqueria SDK like, `getDockerimage(imageName: string, envVarName: string)`
- [ ] Adjust ligo plugin to use this method
- [ ] Adjust archetype plugin to use this method
- [ ] While we're editing relevant code, we may as well adjust any image names to use the latest version tag of the docker image.

### 🔭 Scope

Associated ticket: https://github.com/ecadlabs/taqueria/issues/1425

### Code of Conduct

mweichert:
# 🌮 Taqueria PR

## 🪁 Description

There are four plugins currently which use docker images to facilitate their functionality:
- ligo plugin
- archetype plugin
- flextesa plugin
- tezos-client plugin

🚦 Required to merge:

- [x] 🏄‍♂️ Another team member has built this branch and done manual testing on the change
- [x] 🔱 The test plan has been implemented and verified by an SDET
- [x] 🐚 Required changes to the VScE have been made
- [x] 🪸 Required updates to scaffolds have been made

## 🛩️ Summary of Changes

Users were looking for the ability to override the docker image being used. This is a common request in CICD, and thus we opted to allow overrides using environment variables. This is not mutually exclusive from adding an option later to allow specifying version #s in the config file, such as the ligo version.

## 🎢 Test Plan

- I introduced a new function in the SDK called `getDockerImage`, and added a unit test for it in taqueria-sdk.spec.ts
- For each plugin, I added a test case to assure that each plugin could have a custom image specified using a plugin. To do this, I added a hidden task for each plugin called `get-image` that our tests can use to verify what image is being used. 

## 🛸 Type of Change

- [ ] 🧽 Chore ➾
- [ ] 🛠️ Fix ➾
- [x] ✨ Feature ➾
- [ ] 👷 Refactor ➾
- [ ] 🧪 Pre-Release ➾
- [ ] 🚀 Release ➾


AlirezaHaghshenas:
One general question?
Why do we enable this through environment variables and not by optional settings in `config.json`?

jchenche (commented):
Some e2e tests are failing. Can we address those first?

taqueria-plugin-ligo/common.ts
```diff
@@ -14,7 +15,7 @@ export type IntersectionOpts = CompileOpts & TestOpts;
 type UnionOpts = CompileOpts | TestOpts;
 
 // Points to the latest stable version. Needs to update this as part of our release process
-export const LIGO_DOCKER_IMAGE = 'ligolang/ligo:0.51.0';
+export const LIGO_DOCKER_IMAGE = getDockerImage('ligolang/ligo:0.51.0', 'TAQ_LIGO_IMAGE');
```
jchenche, line 18: Doesn't sound like good practice to have a function return a value that may be different with each invocation and assign it to a global constant.

mweichert:
> One general question? Why do we enable this through environment variables and not by optional settings in `config.json`?

@AlirezaHaghshenas somehow I missed your comment here. It was thought that this would be easiest for CI/CD setups. However, I don't think this is mutually exclusive from also offering the ability override by config. I think that's a great idea too. But this was just meant to resolve a particular use-case.

taqueria-plugin-ligo/common.ts
mweichert: I see this as effectively the same as a constant value - it just can be temporary overwritten for a particular invocation. The value doesn't change over the lifetime of the execution, so I thought this was fine. Did you have something else in mind? I'm open to other approaches.

jchenche: I think it'd be better to use `getDockerImage()` when we are using the image. And then the function would depend on LIGO_DOCKER_IMAGE as a parameter.

e.g. `getDockerImage(LIGO_DOCKER_IMAGE, 'TAQ_LIGO_IMAGE')`

taqueria-plugin-ligo/compile.ts
```diff
@@ -93,7 +93,8 @@ const compileContract = (parsedArgs: Opts, sourceFile: string): Promise<TableRow
 		})
 		.catch(err => {
 			sendErr(`\n=== For ${sourceFile} ===`);
-			sendErr(err.message.replace(/Command failed.+?\n/, ''));
+			sendErr(err.toString());
```
jchenche, line 96: Seems like we are replicating the error message here. Can we revert this?

jchenche (changes requested):
This PR is outdated. Can you merge main in and resolve all conflicts?

taqueria-plugin-ligo/common.ts
```diff
@@ -18,7 +19,7 @@ export type IntersectionOpts = LigoOpts & CompileOpts & TestOpts;
 type UnionOpts = LigoOpts | CompileOpts | TestOpts;
 
 // Points to the latest stable version. Needs to update this as part of our release process
-export const LIGO_DOCKER_IMAGE = 'ligolang/ligo:0.55.0';
+export const LIGO_DOCKER_IMAGE = getDockerImage('ligolang/ligo:0.55.0', 'TAQ_LIGO_IMAGE');
```
jchenche, line 22: It is easier to reason and reduces cognitive overhead to only assign literals to global constants

jchenche (changes requested):

AlirezaHaghshenas (approved):
LGTM.

I would personally be happier if we had not created the get-image tasks.

Is there an end-user use-case for those?

mweichert:
> LGTM.
> 
> I would personally be happier if we had not created the get-image tasks.
> 
> Is there an end-user use-case for those?

It's just to verify what docker image is used by a particular plugin, just in case you need to double check. I believe I made these tasks hidden from the UI.

taqueria-plugin-ligo/common.ts
mweichert: OK, I'll adjust. Thanks!

jchenche (approved):
LGMT!

PR #1466, opened 2022-11-09
✨ Feature ➾ Implement arbitrary command passing from Taq CLI to underlying binary (LIGO and Octez-client)

### LINKED ISSUE #1460: 🌋 E|F ➾ Allow users to invoke arbitrary commands for underlying tools such as LIGO
### Product Board Link

_No response_

### 🚥 Status

- [X] 🔮 Product/Feature Design
- [X] ✍🏻 Architecture/Implementation Planning
- [X] 🏗️ Under Construction
- [x] ⛳ Feature Complete
- [ ] ⛵ GTM / Release

### ✅ To Do

_No response_

### 📐 Feature Description

**Proposal to allow arbitrary commands and options:**

_**Motivation:**_ sometimes LIGO developers want to use a command that’s currently unavailable in our LIGO plugin, such as `ligo compile expression`, and sometimes they want to use a option of a command that is available but doesn’t take such option, like `ligo compile --protocol`. If we allow some plugins such as the LIGO one to have a way of allowing these, then they wouldn’t be blocked by us.

_**The Spec would look like the following:**_ for some plugins (let’s take LIGO), they’ll provide a task aptly named `ligo` that takes a `--command` flag. The value for the flag would be a full LIGO command. UI would look something like this:
`taq ligo --command "compile contract --protocol kathmandu"`

_**Extra benefit:**_ via Google Analytics, we may be able to see which “non-taq task” commands and options (e.g. `taq ligo --command "someCmd someOptions"`) are most frequently used and then we can decide if we want to create a dedicated taq task for it.

### 🚩 Problem or Use Case

_No response_

### ⚖️ Acceptance Criteria

✨ Features: 
- [ ] ... 
 
 📚 Docs: 
- [ ] ... 

### 🔭 Scope

🏎️ In Scope: 
- ... 
 
 🙅‍♂️ Out of Scope: 
- ... 
 
 

### 🔗  Related Issues and Subtasks

_No response_

### 🦄 Sizing

_No response_

### Code of Conduct

jchenche:
# 🌮 Taqueria PR

Fixes #1460 

## 🪁 Description

Allow arbitrary commands to be ran by underlying binary (such as LIGO) from the Taq CLI.

Bonus: allow users to specify Michelson contract files with or without the `.tz` extension.

🚦 Required to merge:

- [x] 🤿 Corresponding changes have been made to all documentation

## 🛩️ Summary of Changes

_Please include a summary of the changes to the codebase._ _Please also include relevant motivation and context for the change_
_(e.g. what was the existing behaviour, why did it break, etc.)_

## 🎢 Test Plan

_Please describe the testing strategy and plan for this PR. Keep this lightweight and anticipate any testing challenges_

## 🛸 Type of Change

- [x] 🧽 Chore ➾
- [x] 🛠️ Fix ➾
- [x] ✨ Feature ➾
- [x] 👷 Refactor ➾
- [ ] 🧪 Pre-Release ➾
- [ ] 🚀 Release ➾


taqueria-plugin-tezos-client/client.ts
```diff
@@ -1,17 +1,42 @@
...
+	return `${baseCmd} ${cmd}`;
+};
+
+const runArbitraryClientCmd = (parsedArgs: Opts, cmd: string): Promise<string> =>
+	getArbitraryClientCmd(parsedArgs, cmd)
+		.then(execCmd)
+		.then(async ({ stdout, stderr }) => {
+			if (stderr.length > 0) sendWarn(stderr);
+			if (stdout.length >= MAX_OUTPUT_LIMIT) {
+				const outputFile = './octez-client.output';
```
mweichert, line 23: What might be best is to send a unique id (hash of microtime + request args might be sufficient) along with the request, and add that as a prefix/postfix to the filename used. This way users can invoke this ligo command with different purposes simultaneously without clobbering the result.

taqueria-plugin-tezos-client/client.ts
```diff
@@ -1,17 +1,42 @@
...
+	const arch = await getArch();
+	const flextesaImage = await getFlextesaImage(arch);
+	const baseCmd =
+		`docker run --rm -v \"${projectDir}\":/project -w /project --platform ${arch} ${flextesaImage} octez-client`;
+	return `${baseCmd} ${cmd}`;
+};
+
+const runArbitraryClientCmd = (parsedArgs: Opts, cmd: string): Promise<string> =>
+	getArbitraryClientCmd(parsedArgs, cmd)
+		.then(execCmd)
```
mweichert, line 19: The real problem here is probably the use of execCmd, which will buffer ALL output.

taqueria-plugin-tezos-client/client.ts
mweichert: At looking closer, although I think having a unique command invocation id passed along with plugin requests could be useful, rather than write to file in this case, we should avoid using execCmd in this case and write another method for the SDK which shares (or at least pipes) output to the main process with minimal buffering.

mweichert (changes requested):
Great idea and solution, Jimmy.

The only thing I'd suggest changing is writing to a file to mitigate the buffering. Instead, use something other than exec to avoid buffering in the first place.

taqueria-sdk/index.ts
```diff
@@ -73,6 +75,21 @@ export const execCmd = (cmd: string): LikeAPromise<StdIO, ExecException> =>
 		});
 	});
 
+export const spawnCmd = (fullCmd: CmdArgEnv): Promise<StdIO> =>
+	new Promise((resolve, reject) => {
```
mweichert, line 79: @jchenche nice to see a spawnCmd method added to the SDK.

However, there are a few problems. 

I think we should first define an intent for this method: "spawnCmd is used to perform non-blocking with minimal (if any) buffering, and do not require independent pipes for stdout/stderr/stdin".

`child_process.spawn` can also be used for launching child processes that have custom processing needs, but I think we should focus our method for the use-case when "independent pipes" aren't required, as developers can always import `child_process.spawn` directly.

After refining the intent of `spawnCmd`, there are some problems we'll find with our implementation:

## spawnSync is used instead of spawn.
This has a few problems in itself:
1) the event loop is blocked. This means that other microtasks, such as any promises waiting to be resolved, or any internal asynchronous microtasks started by the runtime, must wait till spawnSync's child process is closed. Looking at the definition is our function, "waiting" is not permitted here. Concretely, if you look at our use-case, there's no desire for any kind of blocking I/O.

2) Mixing asynchronously and synchronous code is frowned upon as you have to be so conscious not to create incident logic errors.  See this code:
```
const fs = require('fs');
fs.readFile('/file.md', (err, data) => {
  if (err) throw err;
  console.log(data);
});
fs.unlinkSync('/file.md');
```

This kind of thing is hard to catch in a code review, as you have to keep track in your head, "is this using section of the code using async operations or sync operations?". That kind of cognitive load is highly undesirable and unproductive, and thus, its a best practice to use just one or the other, and in the majority of use-cases, async is usually want is wanted.

## We're buffering ALL output from the stdout and stderr pipes:
`child_process.spawn` will by default, create pipes for stdout/stderr/stdin, and have internal buffers used to capture the data microtask: `on.('data', buffer => handler(buffer))`. The data microtask can produce new buffers of data each time IPC data is sent from the child process, meaning there are independent buffers each time there is new data to process, and these buffers also have independent lifecycles so the garbage collector can ran rather optimal.

However, we're calling .toString() on both of the open pipes:
```
resolve({
    stdout: child.stdout ? child.stdout.toString() : '',
    stderr: child.stderr ? child.stderr.toString() : '',
});
```

By calling the .toString() method on a pipe, we're effectively doing the following:
```
return new Promise((resolve, reject) => {
  const stdout = fnWhichSpawnsAndReturnsStdoutPipe();
  const everythingBuffered  = []
  stdout.on('data', data => everythingBuffered.push(data.toString()))
  stdout.on('close', exitCode => exitCode === 0 ? resolve(everythingBuffered.join('')) : reject(exitCode))
});
```

# Recommendation:
- Use `spawn` instead of `spawnSync`.
- `spawnCmd` should resolve with the statusCode, or reject on error:
```
child.on('exit', resolve)
child.on('exit', reject)
```
- I believe we can use the `stdio` configuration option to control what pipes we open. After quickly looking at the docs, I think you can use `'inherit'` as the value which will ensure that the parent and child processes just share the same stdout/stderr/stdin pipes. If that doesn't work, then we just should just pipe the child's stream to the parent.

taqueria-plugin-tezos-client/client.ts
```diff
@@ -1,17 +1,59 @@
...
+	);
+	const args = baseArgs.concat(processedUserArgs);
+	const envVars = {};
+	return [binary, args, envVars];
+};
+
+const runArbitraryClientCmd = (parsedArgs: Opts, cmd: string): Promise<string> =>
+	getArbitraryClientCmd(parsedArgs, cmd)
+		.then(spawnCmd)
+		.then(({ stdout, stderr }) => {
```
mweichert, line 45: See comments below. We shouldn't process stdout or stderr for this use-case.

mweichert (changes requested):
Hey Jimmy,

Unfortunately, this requires some adjustment. I tried to provide as much detail as I could in my comments about what I found problematic, and how to resolve.  If you have any questions,  I'm happy to hop on a call. - although it might be best to read the following first:
https://nodejs.org/en/docs/guides/dont-block-the-event-loop/

Thanks Jimmy!

taqueria-sdk/index.ts
jchenche: Thanks Michael will use spawn.

taqueria-plugin-ligo/ligo.ts
```diff
@@ -1,17 +1,29 @@
...
+	const args = baseArgs.concat(processedUserArgs);
+	const envVars = { 'DOCKER_DEFAULT_PLATFORM': 'linux/amd64' };
+	return [binary, args, envVars];
+};
+
+const runArbitraryLigoCmd = (parsedArgs: Opts, cmd: string): Promise<string> =>
+	getArch()
+		.then(() => getArbitraryLigoCmd(parsedArgs, cmd))
+		.then(spawnCmd)
+		.then(() => `Command "${cmd}" ran successfully by LIGO`)
```
mweichert, line 21: I'd omit this line. Otherwise we'd have to inspect the exit code and determine if indeed it was run successful.

taqueria-plugin-tezos-client/client.ts
```diff
@@ -1,17 +1,53 @@
...
+	);
+	const args = baseArgs.concat(processedUserArgs);
+	const envVars = {};
+	return [binary, args, envVars];
+};
+
+const runArbitraryClientCmd = (parsedArgs: Opts, cmd: string): Promise<string> =>
+	getArbitraryClientCmd(parsedArgs, cmd)
+		.then(spawnCmd)
+		.then(() => `Command "${cmd}" ran successfully by octez-client`)
```
mweichert, line 45: Same with this one. I'd omit the success msg, or if that's required, then to inspect the exit code and ensure its 0 before giving a success msg.

taqueria-sdk/index.ts
```diff
@@ -73,6 +75,18 @@ export const execCmd = (cmd: string): LikeAPromise<StdIO, ExecException> =>
 		});
 	});
 
+export const spawnCmd = (fullCmd: CmdArgEnv) =>
+	new Promise((resolve, reject) => {
+		const cmd = fullCmd[0];
+		const args = fullCmd[1];
+		const envVars = fullCmd[2];
+		const child = spawn(cmd, args, { env: { ...process.env, ...envVars }, stdio: 'inherit' });
+		child.on('close', code => {
```
mweichert, line 84: After seeing your implementation, I see now why you had the success msgs. I was thinking of just doing:
```
child.on('close', resolve); // leave it up to the caller to decide what to do.
child.on('error', reject);
```

The above is technically correct as different programs might use exit codes that have significant meaning, such as https://tldp.org/LDP/abs/html/exitcodes.html

Although atypical, non-zero exit codes doesn't necessarily mean that there was a failure.

mweichert (approved):
Hey Jimmy, I'm going to approve this, but I did add a few more comments. I do think that spawnCmd should return the exit code to the caller and leave it to caller to interpret, as different programs use status codes for different purposes, and non-zero doesn't neccesarily mean a failure.

hu3man (approved):

hu3man:
Looks like just a few merge conflicts left @jchenche but docs and tests are looking great!

PR #1486, opened 2022-11-14
🧽 Chore ➾ (ci) pin node version to v16.16.x

GImbrailo:
# 🌮 Taqueria PR

## 🪁 Description

In CICD we need to pin the `node` version to a v16 LTS version.

## 🛩️ Summary of Changes

Changed all `setup-node` actions in the workflows to pin to `v16.16.x`

## 🛸 Type of Change

- [ ] 🧽 Chore ➾

### DIFF
--- a/.github/workflows/vscode-extension.yml
+++ b/.github/workflows/vscode-extension.yml
@@ -45,7 +45,7 @@ jobs:
       
       - uses: actions/setup-node@v3
         with:
-          node-version: 16.16.x
+          node-version: lts/gallium
           cache: "npm"
 
 


.github/workflows/docker-publish.yml
```diff
@@ -28,6 +28,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
+        with:
+          node-version: 16.16
```
danielelisi, line 32: How about using `lts/gallium` ? This will make sure you're always running on the latest v16 release with security fixes

GImbrailo: Thanks for the tip. Done!

BGMurphy (approved):

danielelisi (approved):

PR #1571, opened 2022-12-02
🛠️ Fix ➾ Clean all docker containers before running any tests

jchenche:
# 🌮 Taqueria PR

Some tests are failing most of the time on Mac Minis but not Linuxes

### DIFF
--- a/bin/before-all-tests.sh
+++ b/bin/before-all-tests.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e # exiting on error
 
-docker ps > /dev/null
+docker ps > /dev/null # making sure docker is installed, running, and the user has permission
 
 # Clean all docker containers before running any tests
 docker rm -vf $(docker ps -aq) > /dev/null 2>&1 || echo "There are no docker containers to remove"
--- a/bin/before-all-tests.sh
+++ b/bin/before-all-tests.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 set -e # exiting on error
 
-docker ps # making sure docker is installed, running, and the user has permission
+# Clean all docker containers before running any tests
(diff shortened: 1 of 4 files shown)


AlirezaHaghshenas (approved):

bin/before-all-tests.sh
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 set -e # exiting on error
 
-docker ps # making sure docker is installed, running, and the user has permission
+docker ps > /dev/null # making sure docker is installed, running, and the user has permission
```
michaelkernaghan, line 4: if docker ps goes to dev null here then the comment should be changed

bin/build-all.sh
```diff
@@ -9,7 +9,7 @@ if [ "$0" == "./bin/build-all.sh" ] && [ -f index.ts ]; then
     echo ""
     echo '**********************************************'
     echo "** Checking Dependencies"
-    docker ps
+    docker ps > /dev/null
```
michaelkernaghan, line 12: if comments on the other file while not in this file too?

michaelkernaghan (commented):
two small comments about comments. otherwise fine :)

bin/before-all-tests.sh
jchenche: @michaelkernaghan changed to what? 🤔 

@AlirezaHaghshenas ideas?

michaelkernaghan: making sure docker is not running

bin/build-all.sh
michaelkernaghan: ok cool

bin/before-all-tests.sh
jchenche: but that's what the comment is saying, and more. It's not inaccurate

jchenche: you mean to say "making sure docker is running"?

michaelkernaghan (approved):

PR #1604, opened 2022-12-09
Config.json schema update

ricklove:
- Auto migrate config.json to the new config file schema (ConfigFileV2).
- Transform the new schema to the runtime Config type (so all current code still works the same).
- Transform the runtime Config type back to the new schema when saving.
- Added to the unit tests to verify this works in every direction.
  - unit tests already had tests for read and writeJsonFile, so seemed the best place
- split the config V2 environment secrets/state into their own files
- add read/writeConfig functions to access the ConfigV2 runtime object (instead of the Config compatibility runtime object)
  - when calling writeConfig it can be specified what should go into main config and config.local.[environment].json
  - there may also need logic to keep things where they are already 
    - i.e. if a field is in the environment config file, keep it there when updated
 - passing:
   - [x] unit
   - [x] integration
   - [x] e2e

---

# 🌮 Taqueria PR

## 🪁 Description

_Explain it to me like I'm 5! Keep this short and descriptive_

🚦 Required to merge:

## 🛩️ Summary of Changes

_Please include a summary of the changes to the codebase._ _Please also include relevant motivation and context for the change_
_(e.g. what was the existing behaviour, why did it break, etc.)_

## 🎢 Test Plan

_Please describe the testing strategy and plan for this PR. Keep this lightweight and anticipate any testing challenges_

## 🛸 Type of Change

- [ ] 🧽 Chore ➾
- [ ] 🛠️ Fix ➾
- [ ] ✨ Feature ➾
- [x] 👷 Refactor ➾
- [ ] 🧪 Pre-Release ➾
- [ ] 🚀 Release ➾


ricklove:
Looks like some of the e2e tests will need to be updated

taqueria-protocol/types-config-files.ts
```diff
@@ -0,0 +1,343 @@
...
+		accounts: !config.accounts
+			? undefined
+			: Object.fromEntries(
+				Object.entries(config.accounts)
+					.map(([k, v]) => [k, { balance: { amount: v, units: `mutez` } }]),
+			),
+		contracts: config.contracts,
+		environmentDefault: config.environment?.default as string,
+		environments: Object.fromEntries(
+			Object.entries(config.environment ?? {})
```
AlirezaHaghshenas, line 111: If a user has a network and sandbox, but only one env is refering to both, they will lose the second.
One solution will be to make a flat array of environments and their content (sandboxes and networks), so that if an env has 2 sandboxes and 3 networks, 5 copies of that env is created. Then we remove duplicates, then we convert it to the current format.

This probably is not going to happen for any user anyway, so the current solution is perfectly valid as well.

ricklove: Yeah, this will lose data in that case, but I couldn't find any case where that is actually used (other than testing). Let me know if you are aware of any config.json in the wild that might be like that.

AlirezaHaghshenas: No, as I said, it's just speculation.

ricklove:
Looks like an e2e test is failing now - checking

ricklove:
Should be good to go

PR #1638, opened 2022-12-15
 🛠️ Fix ➾ Interpretation and parsing of boolean options given to a plugin

mweichert:
# 🌮 Taqueria PR

## 🪁 Description

The taq CLI communicates with plugins by passing options as CLI arguments. In retrospect, I think passing via stdin would have been more appropriate but that's a different issue. :)

In any case, a CLI argument of `--debug 'false'` was being wrongly interpreted, such as the ParsedArgs object would have the `debug` property set to `true` instead of `false`. This PR fixes that.

🚦 Required to merge:

## 🛩️ Summary of Changes

Fixed the logic of the formatArgs() function in the SDK.

### DIFF
--- a/taqueria-sdk/index.ts
+++ b/taqueria-sdk/index.ts
@@ -233,7 +233,8 @@ const formatArgs = (args: Record<string, string>) => {
 	const entries = Object.entries(args).map(
 		([key, value]) => {
 			if (key === 'config') return [key, JSON.parse(value)];
-			if (value === 'false' || value === 'true') return [key, Boolean(value)];
+			else if (value === 'false') return [key, false];
+			else if (value === 'true') return [key, true];
 			return [key, value];
 		},
 	);


taqueria-sdk/index.ts
```diff
@@ -233,7 +233,8 @@ const formatArgs = (args: Record<string, string>) => {
 	const entries = Object.entries(args).map(
 		([key, value]) => {
 			if (key === 'config') return [key, JSON.parse(value)];
-			if (value === 'false' || value === 'true') return [key, Boolean(value)];
+			else if (value === 'false') return [key, false];
+			else if (value === 'true') return [key, true];
```
jchenche, line 237: Isn't
Boolean('false') the same as false
And 
Isn't Boolean('true') the same as true
?

mweichert: Nope - I would have thought so too. Hence the change/fix.

jchenche (approved):

hu3man (approved):

taqueria-sdk/index.ts
egarson: That is really quite disconcering.

Issue #1162, opened 2022-08-20
🚧 Dev Task: [EPIC] Implement Sandbox Baking Options


hu3man:
### 🚥 Definition of Done

- [X] 🎯 Planning and Design
- [ ] 🏎️ Dev Implementation (PR merged)
- [ ] ✍️ Release Notes Added
- [ ] 🏆 Product Acceptance

### ✅ To Do

- [x] Meet with Jev and Dan to understand mempool and endpoint options

1. Implement manual baking task
2. Develop long-lived task that provides bake-on-demand functionality
3. Implement custom block time

Options:
- Manual baking 
- Auto baking

### ⚖️ Acceptance Criteria

✨ Functional: 

- [x] Add a task to bake a block `taq bake`
- [x] Flextesa plugin can have a custom block time specified by the user 
- [ ] Auto-bake (bake a block as soon as an operation is injected and only when operation injected)

### 🔭 Scope

🙅‍♂️ Out of Scope: 
- Work on Taqueria/plugins other than Flextesa
 
### 🧰 Work Description

Flextesa needs to support various sandbox baking options such as bake-on-demand, auto-bake, and a configurable block time. This will improve the developer experience

For auto bake, the options are:

- mempool pull
- proxy

Auto-bake 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mweichert:
Done

PR #1645, opened 2022-12-16
✨ Feature ➾ 1162, Add baking options to flextesa plugin

mweichert:
# 🌮 Taqueria PR

## 🪁 Description

See issue #1162

🚦 Required to merge:

- [x] 🔱 The test plan has been implemented and verified by an SDET
- [x] 🤿 Corresponding changes have been made to all documentation

## 🛩️ Summary of Changes

Added the ability to set baking to "enabled" or "disabled".

In addition, the flextesa plugin is now image agnostic, removing the need for the @taqueria/flextesa-manager package and custom docker image.

### Baking Enabled

To enable baking:
```js
"local": {
    "label": "Local Tezos Sandbox",
    "rpcUrl": "http://localhost:20000",
    "protocol": "PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg",
    "annotations": {
        "baking": "enabled"
    },
   // ...
}
```

Note, when "baking" isn't specified, its set to "enabled" by default.

When baking is enabled, you can also change the block time:
```
"annotations": {
   "baking": "enabled",
   "block_time": 1
}
```

The `block_time` parameter must be an integer and greater than 0.

### Baking Disabled

To disable baking:
```js
"local": {
    "label": "Local Tezos Sandbox",
    "rpcUrl": "http://localhost:20000",
    "protocol": "PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg",
    "annotations": {
        "baking": "disabled"
    },
   // ...
}
```

With baking disabled, you will need to manually bake blocks, which is done with the `taq bake` task.

You have a few options here:
1) Inject a few operations into the mempool and then run `taq bake local` to bake as many blocks as needed (there's only so many operations that can fit in a single block).
2) Inject a single operation into the mempool and then run `taq bake local` to bake that single operation into a block of its own.
3) Run `taq bake --watch` to watch for injected operations into the mempool and bake immediately when an operation is injected.

### Other notable changes
- The flextesa plugin now uses the oxheadalpha image directly, meaning that we don't need to build or maintain our own image nor do we need the `@taqueria/flextesa-manager` package.
- The protocol field has been removed from Network, as it isn't used or needed
- The protocol field is now honoured by the flextesa option, however you only use a value given by `taq show protocols`.
- If you build quite often, use `SKIP_NPM_CI=true  npm run build-all` to avoid reinstalling NPM packages.

## 🎢 Test Plan

Will add E2E coverage.


taqueria-config.ts
```diff
@@ -34,7 +34,6 @@ export const defaultConfig: Config.t = Config.create({
 	sandbox: {
 		local: {
 			label: 'Local Tezos Sandbox',
-			protocol: 'PtKathmankSpLLDALzWw7CGD2j2MtyveTwboEYokqUCP4a1LxMg',
```
jchenche, line 37: We can now start a sandbox and deploy to it without protocol? I recall not being able to do that for sandboxes

taqueria-plugin-flextesa/proxy.ts
```diff
@@ -545,19 +521,22 @@ const importSandboxAccounts = (parsedArgs: ValidOpts) =>
...
 const getDefaultSandboxName = (parsedArgs: Opts) => {
 	return Object.keys(parsedArgs.config.sandbox ?? {}).shift() ?? 'local';
 };
 
+const taskMap: Record<string, (opts: ValidOpts) => Promise<void>> = {
+	'list accounts': listAccountsTask,
+	'show protocols': listProtocolsTask,
+	'list protocols': listProtocolsTask,
+	'start sandbox': startSandboxTask,
+	'start flextesa': startSandboxTask,
```
jchenche, line 533: I thought `yargs` ignores the second word so `start sandbox` and `start flextesa` and `start <anything>` are the same no? When did we fix that?

jchenche (changes requested):

taqueria-config.ts
mweichert: That, that's fixed as part of this PR. If omitted, it will default the latest version supported by the flextesa image, and is represented as the Alpha protocol.

taqueria-plugin-flextesa/proxy.ts
mweichert: `start flextesa` has always been an alias for the `start sandbox` function, and thus, it was technically correct to include them here. 

Also, we don't use `start [words...]` for any other task than `start sandbox`, and thus, the alias will actually work as yargs appears to do the following:

- `start sandbox` is registered as a command in yargs
- You can then type `start [anything]` and yargs will match that against the `start sandbox` task. Thus, `start flextesa` will get matched against `start sandbox`, which is correct! :P

jchenche: Yes but this is quite bad because the second word is in fact meaningless. Since it doesn't cause problem in this particular case, I am ok with moving on with this. We do need to fix this sometime in the future though.

jchenche (approved):

danielelisi (approved):
Changes to Github workflows LGTM :+1:

michaelkernaghan (approved):
I have a concern about the code analysis warning about the path ambiguities, but lets get this done, as it is passing the test that we have

PR #1757, opened 2023-01-13
🛠️ Fix ➾ Added analytics event tracking to vscode

mweichert:
# 🌮 Taqueria PR

## 🪁 Description

Added analytics tracking to vscode

🚦 Required to merge:

## 🛩️ Summary of Changes

- Included the `@taqueria/analytics` library in vscode
- Send the `taq_vscode_loaded` event when the vscode extension is loaded
- Bumped the type definitions for the vscode API to a later version (not the latest, so that we're at least still some-what backwards compatible with earlier versions of vscode)

### DIFF
--- a/taqueria-analytics/index.ts
+++ b/taqueria-analytics/index.ts
@@ -16,7 +16,7 @@ const toRequestURI = () =>
 	`https://www.google-analytics.com/mp/collect?measurement_id=${MEASUREMENT_ID}&api_secret=${API_SECRET}`;
 
 export const inject = ({ taqVersion, taqBuild, fields, getMachineId, operatingSystem, fetch, ...deps }: Deps) => {
-	const events: StoredEvent[] = [];
+	let events: StoredEvent[] = [];
 
 	const toRequestBody = (client_id: string) =>
 		JSON.stringify({
@@ -39,11 +39,15 @@ export const inject = ({ taqVersion, taqBuild, fields, getMachineId, operatingSy
 			},
 		});
 
+	const empty = () => {
+		events = [];
+	};
+
 	const sendEvents = () =>
 		getMachineId()
 			.then(toRequestBody)
 			.then((body: string) => fetch(toRequestURI(), { method: 'POST', body }))
-			.then(noop)
+			.then(empty)
 			.catch(noop);
 
 	const isTrackingAllowed = () => deps.settings.consent === 'opt_in';
--- a/taqueria-vscode-extension/src/lib/helpers.ts
+++ b/taqueria-vscode-extension/src/lib/helpers.ts
@@ -1,3 +1,4 @@
+import * as Analytics from '@taqueria/analytics';
 import loadI18n, { i18n } from '@taqueria/protocol/i18n';
 import * as NonEmptyString from '@taqueria/protocol/NonEmptyString';
 import { TaqError } from '@taqueria/protocol/TaqError';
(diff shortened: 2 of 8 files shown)


taqueria-analytics/index.ts
```diff
@@ -16,12 +16,15 @@ const toRequestURI = () =>
 	`https://www.google-analytics.com/mp/collect?measurement_id=${MEASUREMENT_ID}&api_secret=${API_SECRET}`;
 
 export const inject = ({ taqVersion, taqBuild, fields, getMachineId, operatingSystem, fetch, ...deps }: Deps) => {
-	const events: StoredEvent[] = [];
+	let events: StoredEvent[] = [];
```
jchenche, line 19: why do we need to mutate `events`?

taqueria-analytics/index.ts
```diff
@@ -30,20 +33,25 @@ export const inject = ({ taqVersion, taqBuild, fields, getMachineId, operatingSy
 		events.push({
 			name,
 			params: {
-				...params,
-				...fields,
+				taq_error: false,
```
jchenche, line 36: What if there is an error? Is the error case handled elsewhere?

analytics.ts
```diff
@@ -132,6 +133,9 @@ export const inject = (deps: UsageAnalyticsDeps) => {
 					taqVersion: parsedArgs.setVersion,
 					operatingSystem: machineInfo.os,
 					fetch,
+					fields: {
```
jchenche, line 136: Is `fields` treated specially by Google Analytics? Can we just flatten `taq_ui` instead of having it inside `fields`?

jchenche (commented):
Have you tested it manually and observed it in the Google Analytics website for both CLI and VSCode?

taqueria-vscode-extension/src/lib/helpers.ts
```diff
@@ -263,12 +266,57 @@ export class VsCodeHelper {
...
 
+	async getAnalytics() {
+		if (!this.analytics) {
+			const taqVersion = (await this.getTaqVersion()) || 'unknown';
+			const taqBuild = (await this.getTaqBuild()) || 'unknown';
+
+			this.analytics = Analytics.inject({
+				taqVersion: taqVersion.trim(),
+				taqBuild: taqBuild.trim(),
+				isCICD: false,
```
AlirezaHaghshenas, line 296: If we implement integration/e2e tests for TVsCE, this might need to be changed

AlirezaHaghshenas (approved):

analytics.ts
mweichert: `fields` is a key/value mapping of event fields that you want to include in all events that are added/tracked. Think of them as default or global fields that you want as part of the event payload regardless of what event you sent.

All key/value pairs in fields get merged with the event data in the addEvent function:
https://github.com/ecadlabs/taqueria/blob/vscode-analytics/taqueria-analytics/index.ts#L32

taqueria-analytics/index.ts
mweichert: Yes, this is setting taq_error to false a default value. It will get overridden in params when a consumer of this package calls `addEvent`.

taqueria-vscode-extension/src/lib/helpers.ts
mweichert: Is there something in particular that you're concerned about Ali that warrants adjustment?

taqueria-analytics/index.ts
mweichert: Yes, so that we can empty the array after sending events in the request to GA. The events array is meant to represent the list of events _to_ send.

jchenche (approved):
Awesome enhancements @mweichert LGMT!

PR #1792, opened 2023-01-19
👷 Refactor ➾ Added V2 API for the toolkit

mweichert:
# 🌮 Taqueria PR

## 🪁 Description

Previously, the API of the toolkit included two functions that would only work when given Config V1 objects.

This PR adds a V2 namespace for functions that work with Config V2 objects.

🚦 Required to merge:

## 🛩️ Summary of Changes

- In my last refactor, I introduced a `loadFromEnv` function which has been renamed to `withEnv`, made private, and used internally by two new public functions, `getConfigV1` and `getConfigV2`. These will return a representation of the project's form in either V1 or V2 respectfully, regardless of what version the actual config file is using. This facilitates backwards and frontwards compatibility.
- Introduces the V2 module, which is the start of an API for working with projects using the V2 config. This API is limited to just three functions for now, but we can flesh this out later.

NOTE: this toolkit and the V2 API will be used in the upcoming taco-shop scaffold refactor.

## 🎢 Test Plan

Testing against the taco-shop scaffold.

### DIFF
--- a/taqueria-toolkit/TaqError.ts
+++ b/taqueria-toolkit/TaqError.ts
@@ -1,0 +1,5 @@
+export class TaqError extends Error {
+	public isTaqError = true;
+}
+export const isTaqError = (err: unknown): err is TaqError =>
+	typeof err === 'object' && (err as object).hasOwnProperty('isTaqError');
--- a/taqueria-protocol/scripts/post-ts-to-zod.ts
+++ b/taqueria-protocol/scripts/post-ts-to-zod.ts
@@ -12,6 +12,7 @@ const PASSTHROUGH_TYPES = [
 	'ProxyTaskArgs',
 	'ProxyTemplateArgs',
 	'SanitizedArgs',
+	'ConfigEnvironmentFileV2',
 ];
 
 /**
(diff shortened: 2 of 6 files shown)


taqueria-toolkit/TaqError.ts
```diff
@@ -0,0 +1,5 @@
+export class TaqError extends Error {
+	public isTaqError = true;
+}
+export const isTaqError = (err: unknown): err is TaqError =>
+	typeof err === 'object' && (err as object).hasOwnProperty('isTaqError');
```
egarson, line 5: I would return "isSubclassOf(TaqError)" and you can elide the isTaqError member.
Small fish.

egarson (approved):
LGTM

taqueria-toolkit/TaqError.ts
mweichert: Hmm, is that new? I'm not aware of TypeScript having this keyword or built-in function. I'd be interested to see how that gets transpiled to JS. We should probably make point of bumping our TS versions as right now we're using 4.7.3.

jchenche (approved):

taqueria-toolkit/TaqError.ts
egarson: Right, when I said "whatever the equivalent is in TS" I didn't realize the type system has no support for this construct. So never mind :grin:

